PR #32, opened 2026-05-12
feat(launchers): phase 2 frontend — wizard, RunAgentModal, ActiveSessionsList

nmgarza5:
## Description

Phase 2 of coding-tool-launchers. **Stacked off PR #31** (Phase 1 backend) — merge that first.

**New / changed UI**:

- `RunAgentModal` rewritten from stub to wizard host. Pulls catalog from Phase 1 `GET /api/launchers`, runs helper probe, POSTs `/api/launch`, navigates to the `agentwiki://` URI. Light + dark + responsive (375px).
  - Probe is refreshed when the setup wizard finishes so a stale "helper missing" reading doesn't bounce the user back into the wizard on the next launch.
  - `selectedId` is kept in sync with the current `available_for_launch` slice — if the catalog refreshes while the modal is open and the chosen tool drops off the launchable list, Run is disabled and selection re-points at the first valid tool.
  - Working-dir input refreshes its default whenever the tool changes (until the user types manually); was previously sticky on the first tool's default.
  - CLI badge reports `null` ("CLI: ?") while the helper probe is pending instead of a definite "missing".
- `SetupWizard` — 2-step wizard (multi-select tools → per-tool checklist). Step 2 only requires `helperState.acked` when the user selected at least one `local_cli` tool; in-app-only flows skip the helper gate + install pane.
- `ActiveSessionsList` — live widget on file viewer, polls `/api/agent-sessions` every 5s, shows the user's open sessions for the page with a Close button.
- `/agents` page gains a top-level **Coding tools** section above existing token management. Helper install status badge (R8#1). Refreshes the probe when the wizard closes so the badge isn't stuck on a stale reading.
- Helpers: `ToolCard`, `ToolStatusBadge`, `InstallHelperPane`, `WorkingDirInput`.

**Styling**: new components use **CSS Modules** (`Component.module.css` adjacent to `Component.tsx`) consuming theme tokens as CSS variables (`var(--color-text-primary)` etc). `globals.css` gains `--radius-{xs,sm,md,lg,pill}` so both CSS Modules and `lib/theme.ts` share the same source of truth for radii. CLAUDE.md "Design tokens" seam updated — new components migrate; existing inline-style components stay until their next material edit.

[...]

PR #33, opened 2026-05-12
feat(launchers): phase 3 helper — npm package, Mac URI handler, real claude verification

nmgarza5:
## Description

Phase 3 of coding-tool-launchers. **Stacked off PR #32** (Phase 2 frontend) — merge that first.

**New package**: `packages/agentwiki-launcher/` — Node 18+ TypeScript CLI published as `@agentwiki/launcher`. Single binary entry, four subcommands:

- `agentwiki-launcher set-endpoint <wiki-url>` — pin wiki endpoint to `~/.agentwiki/endpoint.url` (mode 0600). REQUIRED before first run.
- `agentwiki-launcher run <agentwiki://run?...>` — main flow.
- `agentwiki-launcher probe-ack <agentwiki://probe?...>` — posts ack to wiki for frontend helper detection.
- `agentwiki-launcher dispatch <agentwiki://…>` — entry from the macOS .app stub. Routes by URI action so the stub doesn't need to know the difference.

**Core modules**:

- `machine_id.ts` — uuid v4 persisted to `~/.agentwiki/machine.id` (0600).
- `allowed_binaries.ts` — hardcoded `{claude, codex}` allow-list, closes RCE hole.
- `tmpfile.ts` — secure tmpfile helper, mode 0600.
- `endpoint_pin.ts` — URI's `endpoint` param is IGNORED; refuses if doesn't match pinned (handles base/MCP-path normalization).
- `manifest.ts` — TS mirror of backend pydantic; rejects `${token}` in argv, `${first_turn_prompt}` anywhere, `${prompt_file_path}` in resume.
- `interpolate.ts` — bounded substitution; unknown vars throw.
- `mcp_config/{claude_json,codex_toml}.ts` — render MCP config files. claude JSON includes `"type": "http"` (claude rejects schema without it).
- `spawn.ts` — assembles command. `dirhash` = `cwd.replace('/', '-')` (verified against real `~/.claude/projects/`). Appends `unscoped_workdir_argv` from the manifest when `workingDir` is null. Falls back to a per-session scratch dir `~/agent-wiki-runs/<session_id>/` instead of `$HOME` so the agent doesn't touch claude's hard-coded sensitive-file list and doesn't auto-load the user's global `~/.claude/CLAUDE.md`.
- `terminal/darwin.ts` — opens Terminal.app via a `.command` wrapper + `open -a Terminal.app` (LaunchServices). Earlier osascript→Terminal call was silently dropped because TCC refuses `kTCCServiceAppleEvents` for the unsigned osacompile-produced .app; `.command` + `open` routes through LaunchServices with no AppleEvents permission needed. Wrapper trap fires a curl POST to `/api/agent-sessions/<id>/close` on exit so the session row flips to `closed` the moment the terminal tab dies (within the frontend's 5s poll window). Trap body is a named function (`__agentwiki_on_exit`) not a single-quoted inline — earlier inline form let `shellQuote`'d URL/token segments break out of the outer quote and the curl never ran.
- `exchange.ts` — POST `/api/launch/exchange` with machine_id.
- `cli.ts` — `run`: parse URI → check pinned → exchange → build spawn → openInTerminalApp → post spawn-ok beacon. `probe-ack`: send nonce + machine_id to wiki.
- `claude_trust.ts` / `codex_trust.ts` — pre-mark the launch cwd as trusted in `~/.claude.json` / `~/.codex/config.toml` so the workspace-trust dialog doesn't block an unattended agent.
- `codex_mcp_config.ts` — writes the agent-wiki MCP block into `~/.codex/config.toml` with sentinel markers (codex has no `--mcp-config` flag, so per-session block injection is the only way). Uses `bearer_token_env_var = "AGENTWIKI_MCP_TOKEN"`; the env var is exported by the wrapper before launching codex.
- `install/{postinstall,darwin}.ts` — npm postinstall creates `~/Applications/AgentWiki.app` from a compiled AppleScript (not a bash stub — bash stubs can't receive Apple Events for URL-scheme dispatch). Info.plist patched to declare the `agentwiki://` scheme + `lsregister` to register the URL handler.

**Backend additions (small)**:

- `LaunchBlock.unscoped_workdir_argv: list[str]` on the manifest pydantic model + JSON Schema. Helper appends these flags when `workingDir` is null. Literal flags only — `${var}` interpolation is rejected by the validator. claude manifest gets `--permission-mode bypassPermissions`, codex gets `--dangerously-bypass-approvals-and-sandbox`.
- `LauncherCatalogEntry.unscoped_workdir_warning: str | None` — pre-rendered warning string for the catalog response so the UI can show the user what flags will fire before they click Run with an empty workdir.
- `first_turn_prompt_delivery.method: "positional_arg"` enum value. claude + codex manifests switched from the unwired `stdin` method to `positional_arg`; helper appends `promptText` as the final argv element.
- Prompt builder gains an explicit "use the agent-wiki MCP tools (`edit_doc`, `read_doc`); the wiki repo is NOT in your cwd; `WIKI_PATH:` is the arg to those tools, not a local path" clause so launched agents don't write a local file with the wiki path's name.

**Frontend polish landing in this PR** (all small):

- Inline SVG `ToolIcons` (replaced broken `/icons/*.svg` references with React components matching the existing `WikiIcons` pattern). `ToolCard` accepts `kind` instead of `iconUrl`; in_app tools hide the launcher / CLI badges.
- Wizard Done gate loosened — CLI presence-probe isn't shipped in v1 (no localhost helper server). If `claude`/`codex` is missing at spawn time the helper exits with `cli_not_found` and the wiki UI flips the session to `failed`.
- Wiki page body polls `/api/wiki/file?path=…` every 3s while an agent session is active on the page so MCP edits land in the viewer without manual refresh. Suppressed while editing or viewing history.
- `ActiveSessionsList` chip relabelled "external agent sessions" so it doesn't visually conflict with the document-activity (24-hour history) panel.
- `RunAgentModal` busy state explicitly cleared on success path + sessionStorage bounce-back stash for browsers that navigate away on `window.location.href = uri`.

**Real CLI verification done** (claude + codex installed locally). Confirmed:

- claude session-file dir = `~/.claude/projects/<slash-to-dash-cwd>/<uuid>.jsonl`, NOT sha256.
- claude's `--strict-mcp-config` requires `"type": "http"` in the MCP block.
- codex reads MCP servers only from `~/.codex/config.toml` (no `--mcp-config` flag); helper writes a marked block.
- codex needs `bearer_token_env_var` for HTTP MCP, not a `[headers]` sub-table.
- TCC refuses AppleEvents for the unsigned osacompile .app; `.command` + LaunchServices is the workaround.
- Sourcing `~/.zshrc` from the bash wrapper hangs on `onyx-cli`'s interactive init hook → wrapper drops the source.
- Single-quoted trap body breaks on `shellQuote`'d inner segments → trap converted to a named function.
- Multi-byte prompt content with backticks (markdown code spans like `` `onyx-cli ask --json` ``) is logged via node directly, not echoed from bash, to avoid command-substitution.

26/26 unit tests green. `npm run build` produces clean dist.

## How Has This Been Tested?

- `npm test` (helper package): 26/26 green.
- `npm run build` (helper package): clean dist.
- `npm run typecheck` (frontend): clean.
- Backend launcher tests: 95 passed locally (2 pre-existing Redis-env failures unrelated).
- Live end-to-end exercise: real claude + codex launched against running backend, MCP edit_doc tool call lands in wiki repo with `agent_session_id` stamp, close-on-exit beacon flips session to `closed` within the 5s frontend poll, scratch-dir + bypass flags + workspace-trust pre-mark together avoid every prompt that previously blocked an unattended agent.
- 1 codex-driven review iteration on the rebased branch.

### DIFF
--- a/backend/app/api/launchers.py
+++ b/backend/app/api/launchers.py
@@ -105,9 +105,21 @@ def get_catalog(
                 available_for_launch=_entry_available(m),
                 setup_status={"token": has_token},
                 default_working_dir=default_workdir,
+                unscoped_workdir_warning=_unscoped_warning(m),
             )
         )
     return LauncherCatalog(launchers=entries)
+
+
+def _unscoped_warning(m: Manifest) -> str | None:
+    flags = m.launch.unscoped_workdir_argv if m.launch else []
+    if not flags:
+        return None
+    return (
+        f"No directory set — {m.name} will run in a per-session scratch "
+        f"dir (~/agent-wiki-runs/<id>/) with {' '.join(flags)}. Set a "
+        f"directory to scope it to real code."
+    )
 
 
 # --------------------------------------------------------------------------- #
--- a/backend/app/launchers/manifests/claude_code.json
+++ b/backend/app/launchers/manifests/claude_code.json
@@ -13,7 +13,7 @@
   },
   "mcp_config_format": "claude_json",
   "first_turn_prompt_delivery": {
-    "method": "stdin"
+    "method": "positional_arg"
   },
   "launch": {
     "binary": "claude",
@@ -22,7 +22,8 @@
       "AGENTWIKI_SESSION_ID": "${session_id}",
       "AGENTWIKI_ENDPOINT": "${endpoint}"
     },
-    "cwd": "${working_dir}"
+    "cwd": "${working_dir}",
+    "unscoped_workdir_argv": ["--permission-mode", "bypassPermissions"]
   },
   "resume": {
     "binary": "claude",
@@ -36,7 +37,8 @@
     "env": {
       "AGENTWIKI_SESSION_ID": "${session_id}"
     },
-    "cwd": "${working_dir}"
+    "cwd": "${working_dir}",
+    "unscoped_workdir_argv": ["--permission-mode", "bypassPermissions"]
   },
   "session_id_capture": {
     "source": "file_watch",
--- a/backend/app/launchers/manifests/codex.json
+++ b/backend/app/launchers/manifests/codex.json
@@ -13,24 +13,28 @@
   },
   "mcp_config_format": "codex_toml",
   "first_turn_prompt_delivery": {
-    "method": "stdin"
+    "method": "positional_arg"
   },
   "launch": {
     "binary": "codex",
     "argv": [],
     "env": {
       "AGENTWIKI_SESSION_ID": "${session_id}",
-      "AGENTWIKI_ENDPOINT": "${endpoint}"
+      "AGENTWIKI_ENDPOINT": "${endpoint}",
+      "AGENTWIKI_MCP_TOKEN": "${token}"
     },
-    "cwd": "${working_dir}"
+    "cwd": "${working_dir}",
+    "unscoped_workdir_argv": ["--dangerously-bypass-approvals-and-sandbox"]
   },
   "resume": {
     "binary": "codex",
     "argv": ["resume", "${cli_session_id}"],
     "env": {
-      "AGENTWIKI_SESSION_ID": "${session_id}"
+      "AGENTWIKI_SESSION_ID": "${session_id}",
+      "AGENTWIKI_MCP_TOKEN": "${token}"
     },
-    "cwd": "${working_dir}"
+    "cwd": "${working_dir}",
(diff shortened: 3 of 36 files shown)


nmgarza5:
@greptile

PR #114, opened 2026-05-26
prompt(reconciler): precise task rule, block new sections

Bo-Onyx:
## Summary

Two targeted fixes based on analysis of 1,268 post-prompt-change eval samples:

**1. Refine task creation rule**

The previous blanket rule ("don't create tasks unless work is explicitly described") overcorrected — it was blocking legitimate commits where a call summary explicitly assigns a concrete action to a named person already tracked on the page (3 confirmed false negatives). The new rule distinguishes:
- ✅ Source explicitly assigns a concrete action to a named internal person already on the page → valid task
- ❌ Source describes only an entity's state (contact info, deal stage, company profile) without a named internal assignee → not grounds for a task

**2. Block new headings/sections in scope check**

Analysis showed the model was still creating new sections (e.g. "## Tracking OpenAI API spend") from loosely related sources. Adding "do not create new headings or sections" to the scope check prevents this class of bloat.

### DIFF
--- a/backend/app/llm/prompts/ingest_batch_reconciler.system.md
+++ b/backend/app/llm/prompts/ingest_batch_reconciler.system.md
@@ -38,16 +38,17 @@ The page defines what belongs in it — not the source document. Read the
 existing page: what facts does it record, at what level of detail? Only add
 information the page's own structure clearly calls for. New or related content
 that sits at a different level of detail or serves a different purpose does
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
Closing — no longer needed; the reconciler prompt has since moved on and this change is superseded.

PR #146, opened 2026-05-29
feat(evals): ingest pre-filter threshold sweep

nmgarza5:
## Description

[#145](https://github.com/onyx-dot-app/agent-wiki/pull/145) raised `INGEST_BM25_MIN_SCORE` 1.0→20.0 — good change, but the tuning behind it was a **hand-rolled one-off** (`ingest_search_score_analysis.py`): sweep threshold over 3573 labeled samples, BM25-vs-semantic comparison, k1×boost grid search. The eval framework couldn't do any of it.

**The gap:** `ingest_selector` scores the LLM selector at ONE frozen pipeline config. It has no concept of *tuning the retrieval knob* that feeds the LLM — so every threshold/hyperparam decision stays a bespoke script.

**This PR** adds the missing retrieval-sweep primitive:

- `RetrievalSample` schema — (source doc → candidate page, `bm25_score`, human `relevant` label). Scores cached in the dataset → sweep is offline + reproducible.
- `evals.ingest_selector.sweep` — per-threshold `irrelevant_filtered` vs `relevant_retained` curve, then recommends the highest cutoff still retaining ≥ `--min-retained` relevant pages (the production operating point).
- 50-row synthetic seed dataset + 6 unit tests (monotonicity, exact fractions, operating-point selection, unreachable floor).

```
uv run python -m evals.ingest_selector.sweep --min-retained 0.95
```

## Scope / follow-ups (in README)

Deliberately **offline engine only** this PR. Next:
1. `--score-live` mode — re-score the labeled set through `app.ingest.search` against a live OpenSearch cluster (the expensive part #145 did by hand) + Braintrust push to track the operating point. Needs a running cluster → nightly/manual only, can't validate locally, so not in this PR.
2. Build the production dataset from the human-labeled `ingest_eval_samples` corpus (seed here is synthetic).
3. (Optional) semantic-embedding coverage comparison + k1×boost grid, matching the other two analyses in #145.

## Additional Options
- [ ] [Optional] Override Linear Check


nmgarza5:
@greptile review

nmgarza5:
@greptile review

nmgarza5:
On the `_print_curve` print() finding: kept intentionally. It mirrors `reporting.print_summary` (backend/evals/reporting.py:206-213), the established convention for human-facing CLI scorer tables — stdout, not application logging. Only the final JSON line stays on print() per the runner artifact-capture convention. The two real input-guard findings (single-class dataset, empty --thresholds) are fixed in ccdda56.

PR #153, opened 2026-06-01
fix(helm): derive PUBLIC_BASE_URL from ingress.host + bump chart to 0.3.19

nmgarza5:
## Description

**Bug:** every dev-wiki backend/worker pod crashloops on boot:
`pydantic_core.ValidationError: PUBLIC_BASE_URL is required`.

**Cause:** the backend now hard-requires `PUBLIC_BASE_URL` (`config.py`). The chart added the env var in #141 as `required .Values.publicBaseUrl`, but the deploy's values file never sets it — and `Chart.yaml` was left at `0.3.18`, so chart-releaser's `skip_existing` never republished. The deploy kept pulling the stale `0.3.18` chart that has no `PUBLIC_BASE_URL` at all.

**Fix:**
- Derive `PUBLIC_BASE_URL` from `ingress.host` (`https://<host>`), same pattern as `OIDC_REDIRECT_URI`. `publicBaseUrl` still works as an optional override.
- Bump chart `0.3.18` → `0.3.19` so chart-releaser actually publishes to `gh-pages`.

No `cloud-deployment-yamls` values change needed — `dev-wiki` derives `https://dev-wiki.onyx.app` automatically.

`helm template` verified: no override → `https://dev-wiki.onyx.app`; with override → honored.

## How Has This Been Tested?

### DIFF
--- a/deploy/helm/agent-workspace/Chart.yaml
+++ b/deploy/helm/agent-workspace/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: agent-workspace
 description: Self-updating wiki for AI agents (FastAPI + Postgres 17 + Redis + Next.js).
 type: application
-version: 0.3.18
+version: 0.3.19
 appVersion: "0.1.0"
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile review

nmgarza5:
@greptile

PR #159, opened 2026-06-02
fix(launchers): inline-or-defer seed page, drop redundant re-fetch

nmgarza5:
## Description

**Bug:** a launched agent re-fetched the wiki page already injected into its first-turn prompt. The MCP server's standing "search before you begin" instruction drove a `resources/list` + `read_doc` for content the agent already held. Separately, pages over the 256KB cap were truncated but presented to the agent as complete.

**Fix:**
- **Inline** (body fits): embed `<wiki_page>` + tell the agent it's already loaded — don't re-fetch.
- **Defer** (body over cap): omit the body, tell the agent to `read_doc(WIKI_PATH)` for full content. Never ship a truncated body as if whole.
- Dropped the body-budget truncation math; `_compose` now takes a `seed_note`.

Net: zero round-trip for normal pages, no truncation for large ones.

## Additional Options
- [x] Cherry-pick this PR to the latest release branch

### DIFF
--- a/backend/app/launchers/prompt_builder.py
+++ b/backend/app/launchers/prompt_builder.py
@@ -21,12 +21,12 @@ sizes.
 
 from __future__ import annotations
 
-_MAX_PROMPT_BYTES = 256 * 1024
-_TRUNCATION_MARKER = "\n\n[truncated]"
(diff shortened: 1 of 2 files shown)


nmgarza5:
@greptile

PR #132, opened 2026-05-28
feat(evals): add nightly regression gate with per-surface scorer floors

nmgarza5:
## Description

**Bug shape**: prompt PRs can drop `facts_preserved` by 15 points and the only signal is a downward trend in the BT chart — CI stays green. This adds a final step to the nightly job that enforces a per-surface, per-scorer floor.

**Design**:
- New `evals/ci_assert_nightly.py` — reads each `runs/nightly_*.jsonl`, aggregates per-(surface, model) means, fails if below floor.
- Floors anchor to the 2026-05-28 baseline **minus one bucket of slack** — catch real regressions, don't flap on per-run noise.
- Comment trail on each entry records the baseline so future readers know the reasoning.
- Bump the floor in the same PR that lands a sustained improvement.

**Initial floors**:
- `external_agent`: facts_preserved_avg ≥ 0.80, facts_present_avg ≥ 0.85, update_f1 ≥ 0.95, no_touch_compliance ≥ 0.90
- `process_instruction`: trigger_class_match ≥ 0.80, facts_present ≥ 0.85, facts_preserved ≥ 0.85
- `reconcile_document`: trigger_class_match ≥ 0.45 (known-broken, tracked separately), facts_present ≥ 0.80, facts_preserved ≥ 0.85
- `ingest_selector`: f1 ≥ 0.45 (haiku tail), recall ≥ 0.80
- `triggers`: trigger_match_decision ≥ 0.85, no_false_fire_compliance ≥ 0.90

11 unit tests cover pass/fail above/below floor, wiki_updater's sub-surface split (one file → two surfaces), unknown-filename skip, missing/empty file handling, parametrized boundary checks. Smoke (dry-run) keeps its existing `ci_assert_baseline.py` oracle; this new asserter is a separate seam wired into the nightly workflow.

## Additional Options
- [ ] [Tests added]
- [ ] [Type checked]
- [ ] [Frontend reviewed in light mode, dark mode, and responsive design]
- [ ] [I have made corresponding changes to the documentation]
- [ ] [Mobile UI Looks Good]
- [ ] [cherry-pick this PR to the latest release branch once it has been merged]

PR #163, opened 2026-06-02
fix(evals-ci): set PUBLIC_BASE_URL in eval workflows

nmgarza5:
## Description

The 2026-06-02 scheduled **Evals nightly failed** ([run 26800556197](https://github.com/onyx-dot-app/agent-wiki/actions/runs/26800556197/job/79006105568)).

**Root cause:** [#141](https://github.com/onyx-dot-app/agent-wiki/pull/141) (commit ca52326) made `public_base_url` a **required, validated** field in `app/config.py` (`:54`, validator `:64-69`). The eval workflows never set `PUBLIC_BASE_URL`, so every live eval step crashed at `CONFIG` validation → wrote no output files → the regression gate (`ci_assert_nightly`, #132) then failed on the missing surface files.

**Fix:** set `PUBLIC_BASE_URL=http://localhost:3088` in both eval workflows' `env`. Evals never serve browsers or build launcher links, so any valid value satisfies the validator. Smoke is dry-run (didn't crash) but gets the var too for safety against any CONFIG-loading path.

The regression gate worked as designed — it surfaced the breakage loudly instead of a silent green nightly.

## Additional Options
- [ ] [Optional] Override Linear Check


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #168, opened 2026-06-03
fix(db): create public schema

raunakab:
## Summary

- In Docker, `POSTGRES_DB` + the custom Postgres image handle `public` schema and extension setup automatically
- Running locally against a freshly created database, neither is guaranteed — `alembic upgrade head` fails with `psycopg.errors.InvalidSchemaName: no schema has been selected to create in`
- Both migration entry points now run `CREATE SCHEMA IF NOT EXISTS public` and `CREATE EXTENSION IF NOT EXISTS pgmq` before migrations execute:
  - `init_db()` in `app/db/session.py` (app and worker boot)
  - `run_migrations_online()` in `app/db/migrations/env.py` (alembic CLI)

Note: `pg_textsearch` still requires `shared_preload_libraries=pg_textsearch` in `postgresql.conf` — it's a preloaded library, not a `CREATE EXTENSION` extension.

### DIFF
--- a/backend/app/db/migrations/env.py
+++ b/backend/app/db/migrations/env.py
@@ -60,11 +60,10 @@ def run_migrations_online() -> None:
     connectable = create_engine(_db_url, poolclass=pool.NullPool, future=True)
 
     with connectable.connect() as connection:
-        # Ensure the public schema and pgmq extension exist before Alembic
-        # applies any migrations — both are prerequisites for the migration
-        # scripts and are not created automatically by PostgreSQL itself.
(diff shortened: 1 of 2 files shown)


nmgarza5 (approved):
Are we still using the pgmq extension?

raunakab:
@nmgarza5 Good catch. No, the task queue was migrated to Redis Streams. Removed `CREATE EXTENSION IF NOT EXISTS pgmq` from the bootstrap in the latest commit.

PR #170, opened 2026-06-03
refactor(frontend): migrate to Opal tokens + Tailwind, delete theme.ts

raunakab:
## Summary

- **Delete `src/lib/theme.ts`** — all color, radius, and shadow values now come directly from Opal CSS vars (`--text-05`, `--background-tint-03`, `--border-01`, `--border-radius-08`, etc.) or their Tailwind equivalents; no JS token wrapper needed
- **Strip `globals.css` alias block** — removed all `--color-*` and `--radius-*` pass-through vars; only the composite `--shadow-*` vars remain (Opal exposes raw alpha tokens, not full `box-shadow` strings)
- **Replace all `style={{}}` blocks with Tailwind** across 30+ pages and components; dynamic runtime values (pixel sizes, conditional colors) are the only remaining inline styles
- **Replace custom SVG icon wrappers** with direct Opal icon imports at every call site; delete `WikiIcons` and `Chevron` helper functions
- **ChatWidget markdown styling** moved from a global CSS class into `ChatWidget.module.css` via `markdownComponents` + CSS module classes
- **Update `CLAUDE.md`** to reflect the Opal-direct token approach throughout (remove all `theme.ts` references, fix token table, fix modal/input/button seam docs)


CLAUDE.md
```diff
@@ -292,14 +292,16 @@ installed in `app/main.py:_install_error_handlers`. The frontend's
...
 Before declaring a frontend change done, verify it in **both themes** and at
-**both viewport sizes**. The app supports light and dark mode (toggled via
-`data-theme` on `<html>`; tokens live in `src/lib/theme.ts` + `globals.css`)
-and a mobile breakpoint (`useIsMobile()` from `src/lib/viewport.ts`).
-
-- Don't introduce raw hex/rgb/named colors — always go through `color.*` /
-  `shadow.*` from `src/lib/theme.ts`. New shades must be added to both
-  `:root` and `:root[data-theme="dark"]` in `globals.css`.
-- Don't use `background: "white"` (or any literal) — use `color.bg.page`.
+**both viewport sizes**. The app supports light and dark mode (Opal's token
```
jmelahman: not needed here, this file is probably way too big already lol and these should probably go in a `web/STANDARDS.md`, etc.

frontend/src/components/chat/ChatWidget.tsx
```diff
@@ -1,9 +1,11 @@
 "use client";
 
 import { useCallback, useEffect, useRef, useState, type FormEvent } from "react";
-import ReactMarkdown from "react-markdown";
+import ReactMarkdown, { type Components } from "react-markdown";
```
jmelahman, line 4: not needed here, but the Onyx repo publishes a ChatWidget component. We should probably consume that here the same we do Opal

frontend/src/app/app/settings/page.tsx
```diff
@@ -46,14 +45,17 @@ export default function SettingsPage() {
...
         <LoadingSpinner center />
       </main>
     );
   }
 
   return (
-    <main style={{ padding: isMobile ? "16px 12px" : "24px 32px", maxWidth: 720 }}>
+    <main
+      className="max-w-[720px]"
+      style={{ padding: isMobile ? "16px 12px" : "24px 32px" }}
```
jmelahman: optional & this might be for compatibility, but seems there are still some traces of raw `style` which other places converted to css in this change

jmelahman (approved):

PR #191, opened 2026-06-04
fix(events): show timestamps in the user's local timezone

Bo-Onyx:
## Problem

The Events page renders trigger-fire timestamps in **UTC** when no wiki timezone is configured — e.g. a fire at 09:40 local shows as "Jun 04, 2026, 16:40". (`events/page.tsx:14` did `user?.settings.timezone ?? "UTC"`.)

## Fix

Fall back to the **browser's** timezone via the existing `browserTimezone()` helper (`lib/cron.ts`) instead of hardcoded UTC. Precedence is now: explicit wiki-timezone setting → browser local → UTC (the helper's own safe fallback). This matches how the rest of the UI renders times (the triggers page already uses browser-local `toLocaleString()`).

One line of logic; `formatTs` already converts the stored UTC `ts` correctly — it was only the default zone that was wrong.

## Testing

`npm run typecheck` passes. This was the only `settings.timezone ?? "UTC"` site in the codebase.

### DIFF
--- a/frontend/src/app/app/events/page.tsx
+++ b/frontend/src/app/app/events/page.tsx
@@ -4,6 +4,7 @@ import { Button } from "@onyx-ai/opal/components";
 import { LoadingSpinner } from "@/components/common/LoadingSpinner";
 import { PageHeader } from "@/components/common/PageHeader";
 import { useRequireAuth } from "@/lib/auth";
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
Superseded — folded into the combined time-display PR.

PR #196, opened 2026-06-04
feat(home): recent-page actions menu + active folder in directory sidebar

nmgarza5:
## Description

- Extract `WikiItemActionsProvider` + `WikiItemMenu` so the directory sidebar and the recent-pages grid share one contextual menu and one set of overlays (Share / Rename / Move / Copy Link / Launch Agent / Delete).
- **Recent-page card ⋯** now opens that menu — it was previously inert. Wrapped in a `<span>` trigger because OPAL `Button` doesn't forward refs into `Popover.Trigger asChild`.
- **Sidebar ⋯** swaps in over the leading file/folder icon on hover (no extra column).
- **Active folder:** clicking a folder row marks it active (highlighted); new pages/folders are created inside it; the active folder resets when it (or an ancestor) is deleted. State lives in the provider so deletes can clear it.

## How Has This Been Tested?

### DIFF
--- a/frontend/src/components/wiki/WikiHome.module.css
+++ b/frontend/src/components/wiki/WikiHome.module.css
@@ -224,6 +224,11 @@
   padding: 4px 4px 4px 12px;
 }
 
+/* Ref-forwarding wrapper for the OPAL Button inside Popover.Trigger asChild. */
+.menuTrigger {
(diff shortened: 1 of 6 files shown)


nmgarza5:
@greptile

PR #197, opened 2026-06-04
feat(home): AI draft generation, destination picker, and chat-driven live editing

nmgarza5:
## Description

Turns the home "Start writing with AI" input into an end-to-end drafting flow.

- **Generate:** `POST /wiki/generate` runs a one-shot `draft_generator` agent (prompt → full Markdown doc + title), and the input drops you into the New Document composer pre-filled to review and Create.
- **Destination picker:** the composer gets a Folder dropdown (root = "Home") so the doc lands where you choose instead of always at root.
- **Prompt in chat:** `drafting/init` takes an optional `prompt`, so the drafting chat shows your request as the first turn instead of the generic blank prime.
- **Live editing:** chat edit-turns on the unsaved draft route to a stateless `POST /wiki/revise` (current body + instruction → revised body) and write straight back to the editor via a draft "bridge" in the drafting context. Stateless-on-current-body means iterative edits compose. Template / saved-doc chat flows are unchanged.
- **Error surfacing:** added a global `LLMError` handler → clean `{error}` envelope (503/502/429) instead of a raw 500 when the LLM is unconfigured or fails.

New agents are non-conversational `complete()` calls (no tools); tests patch `app.llm.client.complete`.

## How Has This Been Tested?

### DIFF
--- a/backend/app/api/chat.py
+++ b/backend/app/api/chat.py
@@ -368,8 +368,9 @@ async def drafting_init(
         # "Start writing with AI": the prompt is the user's real first turn
         # (shown), preceded by a hidden instruction — a draft for it is already
         # in the editor, so the agent just acknowledges and offers to refine.
+        prompt = req.prompt
         await run_in_threadpool(lambda: _append(_compose_ai_followup_instruction(), hidden=True))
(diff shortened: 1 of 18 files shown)


nmgarza5:
@greptile

PR #207, opened 2026-06-08
fix(chat): restore ChatWidget FAB pending replacement implementation

raunakab:
## Summary

- Restores `ChatWidget.tsx` and `ChatWidget.module.css` which were deleted in `1958e00` as part of the layout refactor
- Adds back the two lines removed from `app/layout.tsx` (import + `<ChatWidget />` render site)

The FAB and its replacement haven't been built yet. This keeps the feature alive in the interim until the replacement is implemented, at which point these files should be deleted again.

## Screenshots + Videos

image


raunakab:
Fixed in 2f70b13 — keyframe moved to `ChatWidget.module.css`, all three bespoke `<button>` elements replaced with Opal `Button`.

PR #228, opened 2026-06-11
feat(llm): support custom OpenAI-compatible providers

### LINKED ISSUE #169: Support custom OpenAI-compatible LLM endpoints
## Summary

It would be very helpful if Agent Wiki could support custom OpenAI-compatible LLM providers, instead of only the currently hardcoded built-in options.

A lot of self-hosted or third-party providers expose an OpenAI-compatible API, but they require a custom `base_url` and often use custom model names. Right now the app appears to support only the predefined OpenAI option, which makes these providers hard or impossible to configure cleanly.

## Desired behavior

Please add a way to configure an OpenAI-compatible provider with at least:

- API key
- custom base URL
- custom model name
- optional custom display label / provider name

Examples:

- OpenRouter
- Deepseek API
- oMLX local API
- company-internal OpenAI-compatible API

Many thanks for your excellent multi-agengt wili system

nmgarza5:
## Description

Closes #169 — adds a fifth provider slot **Custom** for any OpenAI-compatible gateway (OpenRouter, DeepSeek, vLLM, llama.cpp, internal proxies).

- `providers/custom.py`: Chat Completions streaming + tool calls via the openai SDK with configurable `base_url`. API key optional (keyless local routers). Tolerates gateway quirks: usage-only final chunks, `stream_options` rejection (one retry without), tool calls under `finish_reason: "stop"`.
- Migration `0029`: `custom_api_key` / `custom_base_url` / `custom_display_name` on `llm_settings`. Admin PUT keeps the existing keep-vs-clear secret semantics; validates http(s) scheme, strips trailing slash, rejects pasted `/chat/completions` URLs.
- `/llm/available` lists the custom provider under its display label with the admin's free-text model list (hidden until ≥1 model added).
- Admin Language Models page: Custom card — base URL, optional key, display name, add/remove model chips. Ingest selector follows.

Deliberately out of scope (follow-ups): multiple custom providers, test-connection preflight, per-model capability overrides.

## How Has This Been Tested?

### DIFF
--- a/backend/app/api/admin.py
+++ b/backend/app/api/admin.py
@@ -15,6 +15,7 @@ from app.auth import invites
 from app.auth.deps import require_admin
 from app.ingest import settings as ingest_settings
 from app.ingest.settings import IngestSettings
+from app.llm import providers as llm_providers
 from app.llm import settings as llm_settings
 from app.llm.settings import LLMSettings
 from app.models.admin import (
@@ -204,9 +205,6 @@ def _redact(key: str) -> str:
     return f"{key[:4]}…{key[-4:]}"
 
 
-_ALLOWED_PROVIDERS = ("anthropic", "openai", "gemini", "ollama", "custom")
-
-
 def _llm_view(s: LLMSettings) -> LLMView:
     return LLMView(
         provider=s.provider,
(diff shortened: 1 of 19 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #229, opened 2026-06-11
feat(llm): test-connection preflight for every provider

nmgarza5:
## Description

Follow-up to #228, requested in #169's thread — and generalized per review feedback: every provider gets a preflight, not just custom.

- `Provider` protocol gains `test_connection(settings, *, model)`; all five providers implement it via a shared `run_preflight` scaffold (non-fatal models-listing probe + decisive 1-token completion, ~10s timeout, no retries)
- `POST /admin/llm/{provider}/test`: registry lookup (404 unknown), `check_configured` gate (400 with the provider's own message), model fallback request → saved[0] → active
- Diagnostics are redacted: per-probe status, auth mode, base URL where meaningful, translated error messages — never the key
- Admin UI: "Test connection" button on every provider card (shared component)
- 12 per-provider preflight tests + registry-dispatch route tests

## How Has This Been Tested?

### DIFF
--- a/backend/app/api/admin.py
+++ b/backend/app/api/admin.py
@@ -15,17 +15,15 @@ from app.auth import invites
 from app.auth.deps import require_admin
 from app.ingest import settings as ingest_settings
 from app.ingest.settings import IngestSettings
+from app.llm.errors import LLMError
 from app.llm import providers as llm_providers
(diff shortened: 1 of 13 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #230, opened 2026-06-11
fix(security): credential hints never reconstruct secrets; ingest key show-once

nmgarza5:
## Description

Key-masking audit findings (no LLM provider keys were exposed — these are the two gaps found):

**Bug 1:** `_redact` showed first4…last4 for any key >8 chars — a 9-char key exposed 8/9 characters — and `"•"*len` leaked the exact length.
**Fix:** hint only for values ≥16 chars; fixed-width mask below that.

**Bug 2:** `GET /admin/ingest` returned the raw ingest API key on every read (the page's own copy promises "masked after you leave this page").
**Fix:** reads return `api_key_set`/`api_key_hint`; the raw key appears only in the regenerate response — show-once, held in transient component state.

Also adds a **CRITICAL Greptile rule** ("Secrets Never Leave the Server"): raw credential fields in read responses, log/trace/error leakage, and reconstructable hints are blocking findings.

Tests pin all of it (`test_credential_masking.py`).

## How Has This Been Tested?


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #237, opened 2026-06-16
fix(api): set response_model on /api/update-policy routes

Bo-Onyx:
Follow-up to #234 (review comment). The three `/api/update-policy` routes used the return-annotation style only; the rest of the API (e.g. `permissions.py`) consistently sets `response_model=` on every route.

Adds `response_model=UpdatePolicyResponse` to the GET/PUT/DELETE decorators so FastAPI performs response-model validation/filtering (schema, extra-field exclusion) at the framework level, and stays consistent with the codebase convention.

No behavior change beyond framework-level response shaping. `ruff` + `basedpyright` clean.

### DIFF
--- a/backend/app/api/update_policy.py
+++ b/backend/app/api/update_policy.py
@@ -44,7 +44,7 @@ def _build_response(path: str) -> UpdatePolicyResponse:
     )
 
 
-@router.get("/update-policy")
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
Folded into #235.

PR #244, opened 2026-06-16
feat(metrics): Wiki Pages (Auto-update) dashboard card

Bo-Onyx:
Adds a **"Wiki Pages (Auto-update)"** stat to the Grafana Ingest Pipeline row (right of "Wiki Pages"), showing how many wiki pages have ingestion auto-update **enabled**.

### Metric
- New `wiki_pages_auto_update_enabled` gauge via a custom collector (same pattern as `wiki_pages_total`), computed on scrape.
- **Scales — no per-page enumeration.** `update_policy.count_ingest_enabled_pages(total, list_pages_under)` drives off the *small* set of explicitly-disabled scopes: only protected subtrees are listed (new `fts.paths_under(prefix)`), then resolved most-granular-wins via `disabled_paths` so re-enabled children count back as enabled. Pages under no disabled scope are enabled by definition → `enabled = total` in the common case (zero index work).
- The `list_pages_under` dependency is injected, so the computation is unit-tested without the search index.

### Dashboard
`wiki-ingest-dashboard.json`: new stat at `x=6,w=3`; Wiki Updates shifts to `x=9`, Document Update Results to `x=12,w=12`. Row stays contiguous (0–24).
Screenshot 2026-06-16 at 1 55 26 PM

### Tests / verification
- Unit: `count_ingest_enabled_pages` (no policy → total; folder-disabled with a re-enabled child + a disabled page → correct count).
- `ruff` + `basedpyright` clean; JSON validated (unique ids, contiguous row).
- Verified live on Docker: `/metrics` emits `wiki_pages_auto_update_enabled`; adding a disabled page keeps the enabled count flat.

This is the last remaining update-policy item.

### DIFF
--- a/backend/app/db/fts.py
+++ b/backend/app/db/fts.py
@@ -202,6 +202,33 @@ def count_documents() -> int | None:
         log.warning("fts: count_documents failed", exc_info=True)
         return None
 
+
+def paths_under(prefix: str) -> list[str]:
+    """Indexed page paths under a folder ``prefix`` (e.g. ``"team/"``).
+
+    ``path`` is a keyword field, so a prefix term matches the stored value.
+    ``prefix=""`` matches every page. Bounded to the first 10k matches (the
+    realistic ceiling).
+
+    Returns ``[]`` only when OpenSearch isn't configured. A backend error
+    (e.g. the search endpoint is down while ``count`` still answers) **raises**
+    rather than returning ``[]`` — an empty result must mean "genuinely no
+    pages here", so callers can't mistake a partial outage for an empty folder.
+    """
+    client = _get_client()
+    if client is None:
+        return []
+    c: OpenSearch = client  # type: ignore[assignment]
+    resp = c.search(
+        index=_index_name(),
+        body={
+            "size": 10_000,
+            "query": {"prefix": {"path": prefix}},
+            "_source": ["path"],
+        },
+    )
+    return [h["_source"]["path"] for h in resp["hits"]["hits"]]
+
 def delete_document(doc_id: str) -> None:
     """Remove a page from the index.  ``doc_id`` is the path when called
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
Addressed the `count_ingest_enabled_pages` overcount review:

`fts.paths_under` no longer swallows backend errors to `[]` — an empty result now unambiguously means "no pages here". A search-endpoint outage (while `count` still answers) **raises**, so it propagates to `_WikiAutoUpdateCollector.collect()`'s existing `try/except`, which logs a warning and falls back to `total` instead of silently reporting the wiki as fully auto-update-enabled. `[]` is still returned when OpenSearch isn't configured (consistent with `count_documents()` → `None` → `total=0`). Added `test_count_ingest_enabled_pages_propagates_lookup_failure`.

Also folded in the missed `Chart.yaml` bump (0.3.21 → 0.3.22) so the published chart re-syncs the dashboard.

Bo-Onyx:
Found a robustness bug during local testing and fixed it:

Disabling the **root** scope made `wiki_pages_auto_update_enabled` go **negative** (`-1`) when a disabled *page*-scope policy row references a path that isn't in the search index (a stale or pre-index row). `count_ingest_enabled_pages` adds such a page to the subtracted set unconditionally, so root-disable (which also pulls in every indexed page) could subtract more than `total_pages`. Clamped the result to `max(0, …)` with a comment, plus `test_count_ingest_enabled_pages_never_negative`.

Verified locally end-to-end (rebuilt backend, reindexed dev wiki): `wiki_pages_total 7`, `wiki_pages_auto_update_enabled 2` (Features folder + Setup.md disabled → 5 off, cascade working); root-disable now reads `0` not `-1`; round-trips restore correctly.

PR #273, opened 2026-06-22
refactor(craft): run launches on the triggers queue, drop dedicated worker

nmgarza5:
## Description

Craft launches ran on a dedicated `craft` queue with its own worker **pod**. This folds them onto the existing **triggers** queue and deletes the dedicated queue/worker/pod — a whole idle pod for a low-volume, manual feature isn't worth it.

`triggers` (concurrency 4) is the only viable home: `documents` is concurrency 1 (a 10–60s launch would stall doc reconciliation) and `lightweight_maintenance` forbids slow HTTP. The async work itself stays — provisioning can't run in the web request.

- `craft_launch` → `@triggers_queue.task()`; removed the `craft` queue from `queues.py` + `run_worker.py`
- Helm chart back to three workers (chart `0.3.26`); removed `worker-craft` from `docker-compose.yml`
- Supersedes the craft-worker half of #272

## How Has This Been Tested?

ruff + basedpyright clean; `helm lint`/`template` verified (renders 3 workers, no craft); `test_craft_launch` updated to use `triggers_queue` and runs in CI.

## Additional Options

### DIFF
--- a/backend/app/tasks/craft.py
+++ b/backend/app/tasks/craft.py
@@ -37,7 +37,7 @@ from app.onyx.client import (
     OnyxError,
     OnyxUnreachableError,
 )
-from app.tasks.queues import craft_queue
+from app.tasks.queues import triggers_queue
 from app.wiki import git as wiki_git
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


nmgarza5:
@greptile re-review — addressed both comment findings: dropped the removed-design references ("beats a dedicated queue/pod", "rather than warranting a dedicated queue+pod") so the comments now describe only the current placement constraint.

PR #293, opened 2026-06-25
feat(wiki): cap-reached banner on the page view

Bo-Onyx:
**PR 1 of 3** in the cap-enforcement split for *Taming Bad-Behaved Wikis*. (2 = `wiki.auto_update_capped` activity event + card; 3 = the ingest-pipeline sliding-window restriction that does the actual pausing.)

## What
When a page's ingestion auto-updates reach the admin cap (`count_24h >= cap_24h > 0`), the page view shows a **distinct** banner from the too-frequent-update warning:

> ⛔ **Auto-update paused — update limit reached.** This page hit the workspace limit of N auto-updates in 24 hours…

- **No "Review settings" CTA** — the cap is admin-controlled, so the owner can't act on it (unlike the warning, which keeps its CTA).
- **Takes precedence** over the frequency warning when both would apply.
- **No backend change** — reuses the existing `update-health` facts (`count_24h`, `cap_24h`, `can_manage`). Gated to managers, hidden when auto-update is off.

## Note on merge order
The banner says "paused," but the actual pause lands in **PR 3** (the ingest-pipeline restriction). Recommend merging PR 3 before/with this one so the banner isn't aspirational. The copy is otherwise harmless if it merges first (a page at the cap is about to stop updating regardless).

## Test
`bun run typecheck`, prettier clean. Frontend-only.
Screenshot 2026-06-25 at 2 16 30 PM

### DIFF
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -59,7 +59,6 @@ from app.triggers import repo as triggers_repo
 from app.wiki import (
     acl,
     agent_activity,
-    constants as wiki_constants,
     diff as wiki_diff,
     drafts as wiki_drafts,
     filesystem,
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
🚧 **Held as draft — blocked on PR 3 (the ingest-pipeline sliding-window restriction).**

This banner asserts auto-update is *paused* and reports a resume time, which is only true once PR 3 actually excludes over-cap pages from ingestion. Merge order: **PR 3 first, then mark this ready.** Converting to draft so it can't merge ahead of the pause it describes.

PR #301, opened 2026-06-26
feat(wiki): route the blank picker card through the Blank template

Bo-Onyx:
Frontend half of the **Blank template** (backend seed in #300): makes the picker's blank entry start from the seeded **Blank** template, so the empty start carries an update policy (auto-update **off**) like the original Step-1 vision ("empty template, auto-update disabled as default").

## What
- The blank card (home grid + new-doc composer's `TemplateStrip`) routes through the **"Blank"** template when one exists — so picking blank applies that template's policy. Falls back to a plain blank page when no Blank template is seeded (existing installs).
- Filters the Blank template out of the featured row / template strip so it isn't shown twice.
- Keeps the blank card's active-highlight correct when it maps to the Blank template.

## Depends on
#300 (seeds the `Blank` starter template). Without it, the blank card behaves exactly as before (plain blank) — safe to merge in either order.

## Test
`bun run typecheck` + prettier clean. UI-only.

### DIFF
--- a/frontend/src/app/app/wiki/[[...slug]]/page.tsx
+++ b/frontend/src/app/app/wiki/[[...slug]]/page.tsx
@@ -928,6 +928,15 @@ function TemplateStrip({
   const CARD_WIDTH = 200;
   const STEP = CARD_WIDTH + 8; // card width + gap
 
+  // The "Blank" template is the empty-start entry point — route the blank card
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
Folded the blank-template frontend into #300 (single PR for the whole Blank-template feature) per review.

PR #311, opened 2026-07-01
feat(triggers): multi-action trigger shape and dispatch

nmgarza5:
## Description

Foundation for multi-destination triggers. A trigger's single destination becomes a list of actions, and dispatch loops over them.

- **Shape:** `action_json` and the trigger YAML now hold `{"actions": [{type, message, slack_webhook_id}]}`. A back-compat reader loads old single-destination triggers as one action, so live triggers keep firing.
- **Dispatch:** each action renders its own message and delivers to its own destination. `event_log` records the fire, `slack` posts to the action's channel.
- **Migration 0041:** folds the `slack_webhook_id` column into each action, then drops the column.

The HTTP API and agent tools stay single-action for now. Multi-action authoring is the next PR.

## Additional Options

### DIFF
--- a/backend/app/api/triggers.py
+++ b/backend/app/api/triggers.py
@@ -3,16 +3,16 @@ from __future__ import annotations
 
 import logging
 import re
-from typing import Any
+from typing import Any, cast
 
 from fastapi import APIRouter, Depends, HTTPException, Response, status
 
 from app.auth import User, require_can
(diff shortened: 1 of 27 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #313, opened 2026-07-01
feat(coedit): live-rebase — fold inbound agent commits into open sessions

Bo-Onyx:
## What

When a page under a live co-edit session is committed **out of band** — an agent, an ingest job, or a classic `PUT /file` — the session's buffer has drifted from that new HEAD. Instead of letting it surface as a conflict at save time, we **reconcile the commit into the live buffer immediately** and tell participants to refetch: the agent's edit shows up in the editor, and the buffer stays on top of HEAD so the eventual checkpoint has nothing to reconcile.

This is the "frequent rebase" half of co-editing — merge cleanliness comes from *this* direction (small, per-commit merges), not from checkpoint frequency. Merging immediately is a cheap Postgres buffer update; committing to git still happens on the checkpoint cadence.

**Agent edits are not co-edit ops.** An agent's change reconciles into the buffer but never enters the session op stream / `coedit_ops`, and isn't attributed to an agent-as-participant. Participants get a `resync` frame (refetch the buffer at the new version), not an op.

## How

- **Trigger:** `after_doc_write` → `coedit_rebase.on_wiki_commit` enqueues a rebase for any active session on the path whose `base_sha` is behind the new commit. Fires once, in the committing process.
- **Rebase task** (`lightweight_maintenance_queue` — reads git refs + updates the Postgres buffer; no commit, no LLM, so it stays low-latency): 3-way `merge_content(base@base_sha, current@HEAD, incoming=buffer)`.
  - **Clean merge** → `reconcile_onto` CAS-replaces the buffer + advances `base_sha`, then broadcasts `resync`.
  - **Overlap** → enqueue a checkpoint **now** (its AI 3-way merge resolves it, commits, and syncs the merged buffer back) — not deferred to the periodic scan.
- **`coedit.reconcile_onto`** — version-CAS buffer replace + `base_sha` advance; returns whether the buffer changed. No op logged.

## Bug fix included

`checkpoint_session` previously committed the merged result to git but left the live buffer as the pre-merge (human-only) version — so a *follow-up* checkpoint would `merge(base==current)` → re-commit the human-only buffer and **silently drop a folded-in agent edit**. It now syncs the committed content back into the buffer via `reconcile_onto` (and broadcasts `resync`). The checkpoint's own commit passes `trigger_coedit_rebase=False` so it doesn't self-trigger a rebase.

## Conflict philosophy (v1)

The human + live UI is the safety net: we present a *well-merged* result (clean fold, or the checkpoint's AI-merge for overlaps — clean text, never raw `<<<<<<<` markers), and the human fixes anything imperfect by continuing to type. The rare continuous-typing race (a human op landing mid-merge) is best-effort — the CAS misses and the checkpoint merge is the backstop. Perfect concurrent-conflict resolution is the OT/CRDT endgame, deferred.

## Tests

- `test_coedit_rebase.py` (new): `reconcile_onto` (fold without logging a session op, no-op base advance, stale-version CAS miss), engine (clean fold, skip-when-current, overlap→checkpoint, raced→skip), and end-to-end wiring (agent commit through the gateway folds live; checkpoint doesn't self-trigger).
- `test_coedit_checkpoint.py`: buffer-sync / no-drop-agent-edit regression test.
- 75 coedit tests pass; 614 passed / 1 skipped across the broad `commit_and_fan_out` / `after_doc_write` slice; whole-project ruff + basedpyright clean.

No schema change.

## Out of scope

Frontend handling of the `resync` frame; collapse-via-tip-amend (5c); OT/CRDT.


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #314, opened 2026-07-01
refactor(triggers): store action_json as jsonb

nmgarza5:
## Description

Follow-up to #311. Every reader treats `triggers.action_json` as a dict, so store it natively.

- Column flips from `Text` to `JSONB`; repo helpers pass dicts straight through, dropping the `json.dumps`/`json.loads` round-trip.
- Migration `0044` casts existing text rows in place (`::jsonb`), guarded on the live column type since fresh schemas already get jsonb from `create_all`.
- Conversion path covered by a rewind-and-upgrade migration test, same pattern as the `0030` encryption test.

## Additional Options


nmgarza5:
@greptile

PR #316, opened 2026-07-01
fix(triggers): mirror every slack channel into destination_configs at boot

nmgarza5:
## Description

Needed before tonight's deploy.

The boot reconcile only mirrored Slack channels referenced by a legacy trigger. Channels users registered but never attached to a trigger stay in the dead `slack_webhooks` table and vanish from the destinations UI after the registry cutover.

Fix: mirror every stored `slack_webhooks` row into `destination_configs` before rewriting trigger YAML. Idempotent via the existing `from_slack_webhook` marker.

## Additional Options

### DIFF
--- a/backend/app/slack/webhooks.py
+++ b/backend/app/slack/webhooks.py
@@ -38,6 +38,14 @@ def _to_dict(w: SlackWebhook) -> dict[str, Any]:
     }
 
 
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

PR #317, opened 2026-07-01
feat(coedit): collapse same-session checkpoints via git tip-amend (5c)

Bo-Onyx:
## What

A run of checkpoints from a single co-edit session now collapses into **one git commit** instead of stacking one per checkpoint — so we can checkpoint often (fresh HEAD for agents) without exploding history.

Each checkpoint tags its commit with a `Coedit-session: <id>` trailer. The next checkpoint checks the repo tip: if it still carries *this* session's trailer (⇒ nothing else committed since), it **amends the tip in place**; otherwise (first checkpoint of a run, or an agent/other commit landed and broke the run) it commits anew through the existing 3-way-merge gateway, starting a fresh run.

## How

- **`git.amend_head(rel_path, body, message, author, *, expected_head)`** — `git commit --amend` under `commit_lock`, with a **required HEAD CAS**: re-checks `HEAD == expected_head` and raises `GitHeadMovedError` if anything advanced it, so we never amend a commit that isn't ours. Plus `git.head_sha()` (repo tip) and `git.commit_message_of()`.
- **`checkpoint_session`** — amend-vs-new decision via `_tip_is_ours` (trailer on HEAD). On `GitHeadMovedError` (an external commit slipped in under the CAS) it falls back to the new-commit path. Both paths then `rebase_onto(checkpointed=True)` to sync `base_sha`/`checkpointed_version` (same version-CAS + resync as #313). The amend path runs the same `after_doc_write` fan-out a normal commit would.
- **Additive-history relaxation** — amend rewrites the tip SHA, confined to *provisional co-edit tips* (guarded by the trailer + HEAD CAS), documented in the wrapper.

## Coordination (why it's safe)

- `git commit --amend` only touches the branch tip, so amend is used **only when our checkpoint is still HEAD**. Any other commit (agent, another session) makes the tip not-ours → new commit. The HEAD CAS closes the TOCTOU (external commit between the trailer check and the amend → abort → fall back).
- Our checkpoint/amend commits pass `trigger_coedit_rebase=False`, so they never enqueue a live-rebase against our own tip.
- Buffer/`base_sha` still coordinate through the optimistic version-CAS in `rebase_onto` (a raced human op leaves the session dirty for the next checkpoint) and the `is_ancestor` guard from #313. No new locks.

## Scope

In: the collapse mechanism + amend infra. **Out (→ 5d):** deferring trigger fan-out to run-finalization and re-tightening the checkpoint timers — timers stay coarse (5m/15m) and triggers still fire per checkpoint, so nothing regresses here.

## Tests

- `git.amend_head` replaces the tip (commit count unchanged) + enforces the HEAD CAS.
- Consecutive same-session checkpoints collapse to one commit (amend; same count, new SHA, latest content).
- An external commit between checkpoints starts a **new** commit (run boundary) and the merge keeps both edits.
- All existing checkpoint tests still pass (86 coedit+git green; ruff + basedpyright clean).

*(Note: some `test_wiki_drafts` tests fail locally with a 503 = LLM-not-configured; verified they fail identically on `main` with this branch stashed — a local-env artifact, unrelated to this change.)*

### DIFF
--- a/backend/app/wiki/coedit_checkpoint.py
+++ b/backend/app/wiki/coedit_checkpoint.py
@@ -34,7 +34,7 @@ log = logging.getLogger(__name__)
 def _tip_is_ours(head_sha: str, session_id: int) -> bool:
     """True if the repo tip is *this* session's own checkpoint commit — it
     carries our ``Coedit-session`` trailer, which means nothing else has
-    committed since, so the next checkpoint can amend it in place (5c)."""
+    committed since, so the next checkpoint can amend it in place."""
     return f"Coedit-session: {session_id}" in wiki_git.commit_message_of(head_sha).splitlines()
 
 
@@ -59,7 +59,7 @@ def _commit_message(session_id: int, *, primary_author_id: str | None) -> str:
         if u is not None:
             trailers.append(f"Co-authored-by: {u['name'] or u['email']} <{u['email']}>")
     # Tag the commit with the session id so the next checkpoint recognizes this
-    # tip as ours and amends it in place (collapse) rather than stacking (5c).
+    # tip as ours and amends it in place (collapse) rather than stacking.
     trailers.append(f"Coedit-session: {session_id}")
     lines.append("")
     lines.extend(trailers)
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
Superseded. Working through the org-scale case (one wiki = one linear branch), amend-on-`main` only collapses in a quiet repo — continuous org-wide commits break the run. The correct model needs no amend and no branch: the **Postgres buffer is the live layer**, live-rebase (#313) keeps it reconciled, and `main` receives a single commit **at session close**. The one real downside (agents don't see the live buffer) is solved at the read path — a session-aware `read_doc` that serves the buffer on demand — not by committing mid-session. See design/Co-Editing.md.

PR #320, opened 2026-07-02
fix(tests): freeze the clock in the coedit legacy-timestamp due test

nmgarza5:
## Description

Bug: `test_legacy_space_separated_created_at_normalized_not_overdue` fails for every PR during the first hour of the UTC day. The space-vs-`T` lexicographic comparison it relies on only decides when the overdue cutoff shares the row's date prefix; between 00:00 and 01:00 UTC, `now − interval` lands on the previous date, the day digit wins the comparison first, and the seeded bug can't reproduce.

Fix: freeze `coedit._now` at a fixed mid-day instant. Every stamp and cutoff derives from that one clock, so the assertion is carried by the intended overdue arm at any wall time. Verified green at 00:22 UTC, inside the window that previously failed.

## Additional Options

### DIFF
--- a/backend/tests/test_coedit_repo.py
+++ b/backend/tests/test_coedit_repo.py
@@ -185,10 +185,18 @@ def test_due_includes_overdue_active_session(users):
     assert s.id in _due_ids(idle_seconds=3600, max_interval_seconds=0)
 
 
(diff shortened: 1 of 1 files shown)


nmgarza5:
Superseded by #322, same fix.

PR #324, opened 2026-07-02
feat(slack): bot posting, DMs, and the channel picker

nmgarza5:
## Description

Trigger fires now deliver through the Agent Wiki bot.

- A slack destination config carries one of three targets: the legacy webhook secret, a `channel_id`, or `dm: true`. Dispatch discriminates on shape — webhook posts as before; bot targets resolve the owner's connection token and post via `chat.postMessage` (`conversations.open` for DMs), appending a source line naming the doc.
- Every missing precondition (no target, no connection, undecryptable token, Slack error) logs and degrades to recorded-only; the fire event is always written first.
- `GET /api/connectors/slack/channels` lists postable channels off the caller's own token, for the picker.
- Creating a targetless slack config is rejected at the repo, so undeliverable configs can't exist.

Picker/multi-target UI comes with the Connectors settings work.

## Additional Options

### DIFF
--- a/backend/app/api/slack_connect.py
+++ b/backend/app/api/slack_connect.py
@@ -26,6 +26,16 @@ log = logging.getLogger(__name__)
 _BOT_SCOPES = "chat:write,chat:write.public,channels:read,groups:read,im:write,users:read,channels:join"
 
 
+class SlackChannel(BaseModel):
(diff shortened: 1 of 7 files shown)


nmgarza5:
@greptile

PR #325, opened 2026-07-02
feat(slack): connect button, channel picker, and DM destinations in the UI

nmgarza5:
## Description

Frontend slice that makes the Slack track clickable end to end.

- The Destinations card gains a Slack connection strip: Connect button (driven by the status endpoint's `connect_url`, bouncing back to `/app/triggers`), or team name + Disconnect when linked. Hidden until an admin saves the app credentials.
- Three add flows: pick a channel from the bot's picker endpoint, one-click "DM me", or the existing webhook form. All funnel through the same config-create call.
- List rows label bot vs webhook targets from `has_secret`.

Verified: tsc + prettier clean; the card renders behind auth, so the visual pass is a one-click check at `/app/triggers` on a signed-in session (local `next dev` serves this branch).

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #326, opened 2026-07-02
fix(triggers): minute-level schedule scan, no doc-name dup, cap botocore logging

nmgarza5:
## Description

Three trigger-pipeline issues found while live-testing Slack delivery.

- Schedule triggers fired up to 5 minutes late: the scan cron ran on a `*/5` grid, so any trigger whose cron minute was off-grid waited for the next tick. The scan now runs every minute. A no-op pass is one indexed query.
- Fired messages duplicated the doc name (LLM-added title plus our appended source line). Both render prompts now state that delivery appends an attribution line naming the document, so the model stops adding its own title.
- `LOG_LEVEL=DEBUG` put the root logger at DEBUG, and botocore then logged every stream chunk and full request headers, bearer token included. botocore/boto3/urllib3 are now capped at INFO under any level, with a test pinning the cap.

## Additional Options

### DIFF
--- a/backend/app/tasks/periodic.py
+++ b/backend/app/tasks/periodic.py
@@ -29,7 +29,9 @@ from app.tasks.triggers import evaluate_due_schedule_triggers
 log = logging.getLogger(__name__)
 
 
-@triggers_queue.periodic_task(crontab(minute="*/5"))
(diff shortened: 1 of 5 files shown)


nmgarza5:
Folded into #325.

PR #327, opened 2026-07-02
chore(slack): retire the slack_webhooks store

nmgarza5:
## Description

Slack channels live in `destination_configs` (webhook-secret, channel, or DM targets); the standalone webhook store has no remaining readers.

- Migration `0047` copies any not-yet-mirrored webhook row into `destination_configs` in SQL before dropping the table. The encrypted secret is unbound AES-GCM bytea, so it moves verbatim, and the mirror id derives from the webhook id, so re-runs are idempotent. No upgrade path loses channels.
- The boot reconcile resolves a legacy `slack_webhook_id` through its mirrored config's `from_slack_webhook` marker; only a reference with no mirror anywhere degrades to event-log, under a warning naming the file, owner, and id.
- Removes `app/slack/webhooks.py`, the `SlackWebhook` model, and their tests. A rewind-and-upgrade migration test proves the secret survives the SQL copy and decrypts.

## Additional Options


nmgarza5:
@greptile

PR #328, opened 2026-07-02
fix(triggers): scope trigger payloads to the trigger's docs

nmgarza5:
## Description

Bug: a scheduled trigger scoped to `Individual Notes/Nik Standup.md` kept reporting the page as unreadable on dev-wiki, despite the page having content.

Cause: every trigger payload led with a whole-wiki snapshot capped at 200k chars, assembled in `git ls-files` order. On a large wiki the alphabetically-earlier docs exhaust the budget, so the scoped doc never reaches the LLM — the trigger's scope was only a label, not a filter.

Fix: payloads now carry only the docs under the trigger's scope, on all three paths (delta, new-file-in-dir, schedule). A `.md` scope is that doc, a folder scope is the docs under it, an empty scope remains the budgeted whole wiki. Scoping reuses the same `_in_scope` predicate as the changes filter, the fan-out builds one scope block per distinct scope per commit, and the prompts describe the `SCOPED DOCS` block. A regression test pins that a scoped doc survives an arbitrarily large alphabetically-earlier sibling.

## Additional Options

### DIFF
--- a/backend/app/tasks/triggers.py
+++ b/backend/app/tasks/triggers.py
@@ -93,20 +93,34 @@ def fan_out_trigger_eval(
     after = _read_at(sha, doc_path)
     before = _read_at(f"{sha}^", doc_path)
 
-    # Build the wiki snapshot once and reuse it for every trigger on this
-    # commit. Each trigger pays its own per-call eval (and render-on-match).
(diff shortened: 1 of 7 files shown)


nmgarza5:
@greptile

PR #329, opened 2026-07-02
fix(slack): render markdown list markers as real bullets, suppress link unfurls

nmgarza5:
## Description

Two formatting gaps in Slack-bound trigger messages.

- Slack's mrkdwn has no list syntax, so markdown `- item` lines posted as literal dashes. The converter now turns leading `-`, `*`, `+` markers into `•` with indentation preserved; bold-at-line-start is untouched (the marker rule requires trailing whitespace).
- `chat.postMessage` unfurled every converted link, growing preview cards under notifications. Posts now set `unfurl_links`/`unfurl_media` false.

## Additional Options


nmgarza5:
@greptile

PR #330, opened 2026-07-02
feat(coedit): route human editing through a co-edit session (frontend)

Bo-Onyx:
## What

Cuts `FileViewer` over to **Model A**: all human editing goes through a live co-edit session (frontend half of the "Edit = session" arc).

- **Edit** → joins the page's co-edit session (`useCoeditSession`); the `<textarea>` binds to the shared session buffer. The buffer is seeded with the committed body immediately so there's no blank flash during the join.
- **Save** → checkpoints the buffer to git + leaves the session (`coedit.save()`), then handles rename via `/wiki/move`.
- **Cancel** → `coedit.discard()`: resets the shared buffer to the committed body (leave-time checkpoint is a no-op) + leaves.

## Removed (edit path no longer needs them)

Live convergence eliminates human-vs-human conflicts and the per-user private draft:
- the 5s `/file/autosave` debounce effect
- the resume banner (+ `/file/autosave/rebase`)
- the inline conflict panel + `onKeepMine` / `onUseCurrent` / `onAiConsolidate` (+ `/file/merge`) — human-vs-agent is reconciled by the checkpoint's 3-way merge + a `resync`
- the orphaned `draft` state, `ConflictState`/`DraftResponse` types, and the `diffLines`/`ApiError` imports they used

Edit is hidden while viewing an older commit (editing always targets current HEAD). The not-editing SWR poll still surfaces MCP/agent edits.

## Depends on #331 (merged)

Reads are **session-aware** (#331): while a session is open, `GET /wiki/file` serves the live buffer, so a Save shows immediately without waiting on the async checkpoint commit. This PR relies on that.

## Fixes found during local testing (later commits)

- **Show saved text immediately** — `onSave` sets the body optimistically; `onEnd` no longer refetches the pre-commit HEAD (which raced and flashed stale content). SWR reconciles against the session-aware read.
- **Save during join latency** — the buffer is seeded with the committed body on Edit so `bodyChanged` is accurate before the join resolves (dropped the `coedit.active` gate); pre-join edits are preserved + flushed; `save()` awaits the in-flight join. Also documented that a rename-only save is a no-op checkpoint server-side (`version == checkpointed_version`), so no empty commit.
- **`pump()` self-poisoning (ops silently dropped)** — `pumpPromise.current = run` was assigned *after* invoking the async drain, but a drain with no pending change resolves synchronously and its inline `finally` cleared the handle first, stranding a resolved promise in `pumpPromise` and making every later pump a no-op. This dropped all ops after the first flush — most visibly on **re-Edit** (sessions stuck at `version 0`, edits never committed). Fixed by setting the handle before attaching completion and clearing it via `run.finally(() => { if (current === run) current = null })`.

## Test

- `bun run typecheck` clean.
- Manual + headless (CDP) cycle on the local stack: Edit → type → Save shows immediately; **re-Edit → type → Save** round-trips correctly (ops fire on every edit, both edits persist); rename-only save produces no empty commit; two tabs converge live.

## Not in this PR

- Backend draft endpoints (`/file/autosave*`, `/file/merge`, `git.py` draft helpers) are retired in the **PR3** follow-up.
- Remote carets / selection / typing (CodeMirror) come in **PR4+**; this is the textarea MVP.

### DIFF
--- a/frontend/src/app/app/wiki/[[...slug]]/page.tsx
+++ b/frontend/src/app/app/wiki/[[...slug]]/page.tsx
@@ -10,7 +10,6 @@ import {
   type FormEvent,
 } from "react";
 import { createPortal } from "react-dom";
-import { diffLines } from "diff";
 import ReactMarkdown from "react-markdown";
 import remarkGfm from "remark-gfm";
 import useSWR from "swr";
@@ -56,7 +55,7 @@ import { ShareDialog } from "@/components/wiki/ShareDialog";
 import { CommentsPanel } from "@/components/wiki/CommentsPanel";
 import { UpdateHealthBanner } from "@/components/wiki/UpdateHealthBanner";
 import { UpdatePolicyPanel } from "@/components/wiki/UpdatePolicyPanel";
-import { apiFetch, ApiError } from "@/lib/api";
+import { apiFetch } from "@/lib/api";
 import { listComments } from "@/lib/comments";
 import {
   paintCommentHighlights,
@@ -66,7 +65,8 @@ import {
 } from "@/lib/commentAnchor";
 import { rehypeSourcePos } from "@/lib/rehypeSourcePos";
 import { remarkBareSpaceLinks } from "@/lib/remarkBareSpaceLinks";
-import { useRequireAuth } from "@/lib/auth";
+import { useAuth, useRequireAuth } from "@/lib/auth";
+import { useCoeditSession } from "@/lib/useCoeditSession";
 import {
   useHeaderActionsHost,
   useRightPanelHost,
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
Rebased: dropped the standalone optimistic-guard approach in favor of the **session-aware read** in #331 (serve the live buffer so a save shows immediately without waiting on the async checkpoint commit). **#331 should land before this PR.**

Bo-Onyx:
Pushed fixes + rebased on the merged #331. Summary of the review threads and a bug found during local testing:

- **Rename-only save → empty commit?** No — `checkpoint_session` returns early when `version == checkpointed_version` (both default 0), so a rename-only save (no ops) never commits before `/wiki/move`. Documented inline in `onSave`.
- **Save disabled during join latency** — fixed: the hook seeds its buffer with `committedBody` on Edit (before join resolves), so `bodyChanged` is accurate; dropped the `coedit.active` gate. Pre-join edits are preserved + flushed, and `save()` awaits the in-flight join.
- **`pump()` self-poisoning (found while testing)** — a drain with no pending change resolved synchronously and its inline `finally` cleared `pumpPromise` before the assignment ran, stranding a resolved promise and making every later pump a no-op → ops silently dropped after the first flush (worst on re-Edit, sessions stuck at `version 0`). Fixed by assigning the handle before attaching completion and clearing via `run.finally(() => { if (current === run) current = null })`.

Verified locally (manual + headless): Edit→Save shows immediately, re-Edit→Save round-trips, rename-only makes no empty commit.

PR #333, opened 2026-07-02
feat(coedit): live co-editing signals in the textarea editor

Bo-Onyx:
## What

A middle step toward the rich editor: while editing, surface **who else is co-editing and who's typing**, without waiting on CodeMirror. Remote carets can't be drawn inside a plain `<textarea>` (that's the CodeMirror piece), but the rest of the "Google-Docs feel" signals are here — and remote edits already merge live via the shared buffer.

- **Presence bar** — "Ada, Bo also editing" from the session roster (`coedit.participants`), shown while editing; renders nothing when you're alone.
- **Live "typing…"** per peer, from `cursor` frames. The textarea reports its caret/selection on edit (typing) and on selection change (idle) through a new hook method `reportSelection`, throttled + coalesced (~80ms) with a trailing "stopped typing" ping. Peer badges expire (~4s) if pings go silent (crash/lost tab).
- **Remote edits** already splice into the shared buffer (inbound `op` frames), so co-editors see each other's text with no new work here.

## Hook (`useCoeditSession`)

- New `typing: string[]` (peer user_ids currently typing) and `reportSelection(anchor, head, typing)`.
- Sends own cursor/typing via `POST /coedit/cursor` (throttled); tracks inbound peer typing with per-peer expiry timers; all cleared in `stop()`.

## Not in this PR

Remote **caret/selection rendering** and follow-user — those need CodeMirror (the offsets are already on the wire via `cursor` frames, so it's render-only work later). Client-side pending-op rebase on resync also lands with CodeMirror.

## Test

- `bun run typecheck` clean.
- Two-participant headless run: B joining shows "Peer Bee also editing"; B typing → "Peer Bee typing… also editing"; B stops → badge clears; A typing broadcasts `cursor` + `op`. All verified.

### DIFF
--- a/frontend/src/app/app/wiki/[[...slug]]/page.tsx
+++ b/frontend/src/app/app/wiki/[[...slug]]/page.tsx
@@ -66,6 +66,7 @@ import {
 import { rehypeSourcePos } from "@/lib/rehypeSourcePos";
 import { remarkBareSpaceLinks } from "@/lib/remarkBareSpaceLinks";
 import { useAuth, useRequireAuth } from "@/lib/auth";
+import type { CoeditParticipant } from "@/lib/coedit";
 import { useCoeditSession } from "@/lib/useCoeditSession";
 import {
   useHeaderActionsHost,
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
Fixed the onSelect-clobbers-typing bug (thanks — good catch). Rather than a `justEdited` ref in the component, I fixed it at the source in the hook: `reportSelection`'s third arg is now **`isEdit`** instead of a raw typing flag. An edit marks typing + (re)arms the 1500ms auto-clear; a caret move reports position with the *current* typing state (derived from whether the idle timer is still pending), so it can't clear a live edit. The idle timer is the only path that emits `typing=false`.

Verified in a headless run — a rapid keystroke burst now sends `typing=true` throughout, with a single `typing=false` ~1.5s after the last key (previously it alternated true/false every ~80ms).

PR #334, opened 2026-07-02
feat(coedit): CodeMirror editor with remote carets + selections

Bo-Onyx:
## What

Swaps the edit-mode `<textarea>` for **CodeMirror 6** and renders peers' live **carets and selection highlights** — the Google-Docs feel. Entirely within edit mode; view/render mode is unchanged.

- **`CoeditEditor`** (new, `src/components/coedit/`): a CodeMirror 6 wrapper with the same controlled contract (`value` + `onChange`), markdown syntax highlighting, line wrapping, and local undo/redo. Reports the local caret/selection via `onSelectionChange(anchor, head, isEdit)` — edit vs. caret-move, matching the hook's `reportSelection`.
- **Remote carets/selections** from `cursor` frames: `useCoeditSession` now exposes `peers` (positions, pruned when someone leaves). The editor draws a colored, name-labeled caret per peer and a translucent selection range; each peer's color is stable (id hash).
- **Minimal-diff buffer sync**: external buffer changes (remote ops / resync / discard) apply the smallest change rather than a whole-doc replace, so the local caret maps through them instead of jumping to the end.
- **Clamp-safe decorations**: peer decorations rebuild + clamp on every doc change, so a slightly-stale peer offset can't land out of range.

## Scope

This is the render half. **Not** in this PR (next one): remote caret position *mapping between frames* and **pending-op rebase on resync** so un-acked local edits never revert. The presence bar + "typing…" from #333 stay as-is.

## Test

- `bun run typecheck` clean.
- Two-participant headless runs: peer caret ("Peer Bee") + selection highlight render and track edits; local typing sends ops; selecting-all + deleting while a peer's caret is at offset 18 clamps to 0 with **0 console errors/exceptions**.

Deps added: `@codemirror/state`, `@codemirror/view`, `@codemirror/commands`, `@codemirror/lang-markdown`.

### DIFF
--- a/frontend/bun.lock
+++ b/frontend/bun.lock
@@ -5,6 +5,10 @@
     "": {
       "name": "agent-wiki-frontend",
       "dependencies": {
+        "@codemirror/commands": "^6",
+        "@codemirror/lang-markdown": "^6",
+        "@codemirror/state": "^6",
+        "@codemirror/view": "^6",
         "@dnd-kit/core": "^6.1.0",
         "@dnd-kit/modifiers": "^7.0.0",
         "@dnd-kit/sortable": "^8.0.0",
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
Pushed two follow-ups on this branch:

**Review fixes** (`c379a49`): annotate the external buffer-sync dispatch as `Transaction.remote` and skip remote transactions in the updateListener (was echoing remote ops back + falsely flagging local user as "typing…"); reword the module docstring to current-behavior only; `#fff` → `var(--text-inverse, #fff)`.

**Exit UX** (`7b61c06`): collapse the editor exit to a single non-destructive **"Done"** and remove **Cancel**. Cancel was destructive in a shared session — it reset the shared buffer and broadcast it, discarding *everyone's* in-progress edits. In Model A the buffer is the live shared doc, so there's nothing safe to discard. Done checkpoints + leaves; the navigate-away "discard unsaved changes" guard is gone too (leaving loses nothing — the buffer is durable and commits at session close).

Verified: peer op applies with zero false-typing frames; only "Done" shows while editing; type → Done commits + exits.

PR #340, opened 2026-07-03
feat(triggers): reject scopes that don't exist in the wiki, with a did-you-mean hint

nmgarza5:
## Description

Two triggers today were scoped to paths that don't exist (`fun poem.md` vs `fun_poem.md`, `Individual Notes / Nik Standup.md` with spaces). Since #328, payloads carry only the docs under the scope, so a typo'd scope renders "no content found" fires with nothing to catch it.

- Authoring never blocks: watch-for-creation is a supported flow, and separator-insensitive matching false-positives on legitimate names (`todo.md` vs `to-do.md`).
- Instead, create/update responses (HTTP API and agent tools, one shared builder) carry a `scope_warning` when the scope has no watchable docs yet. When the scope near-misses an existing doc, the warning names it ("did you mean `fun_poem.md`?") — and the hint is offered only when the caller can read that doc, so the warning can't be used as an existence oracle for private paths.
- `scope_exists` counts only `.md` docs: a folder holding just a trigger's own YAML has nothing to watch.

## Additional Options

### DIFF
--- a/backend/app/api/triggers.py
+++ b/backend/app/api/triggers.py
@@ -56,10 +56,10 @@ def _config_view(row: dict[str, Any]) -> DestinationConfigView:
     )
 
 
-def _normalize_scope_path(raw: str) -> str:
+def _normalize_scope_path(raw: str, user: User) -> str:
     if not raw.strip():
         raise ValueError("scope_path is required")
-    return triggers_storage.normalize_scope_path(raw)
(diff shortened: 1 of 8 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #345, opened 2026-07-06
feat(triggers): recent-fires API (GET /api/triggers/fires)

nmgarza5:
## Description

Adds `GET /api/triggers/fires`, serving the `trigger.fire` events rows for the caller's triggers. A window `row_number` caps results per trigger (default 3) so one busy trigger can't starve the rest of the Triggers page; passing `trigger_id` narrows to a single trigger and uses the full `limit` instead, which is the history-flyout query. Payloads are flattened into a typed fire view (doc path, change kind, reason, message, destination type) with malformed payload rows degrading to empty fields.

Reuses the events rows `_record_fire` already writes — no new table. Ownership scoping mirrors `/api/events`. The route is declared ahead of the `/{trigger_id}` matchers so `fires` can't be captured as an id. Includes the typed frontend accessor (`getTriggerFires`) for the upcoming Triggers page redesign, which consumes this per the Figma mocks.

## How Has This Been Tested?

New `test_triggers_fires_api.py`: per-trigger cap + ordering, owner scoping, `trigger_id` limit semantics, payload flattening. Live smoke on dev: the route resolves and gates.

## Additional Options

### DIFF
--- a/backend/app/api/triggers.py
+++ b/backend/app/api/triggers.py
@@ -1,11 +1,16 @@
 """FastAPI port of ``app/api/triggers.py`` (Phase 3)."""
 from __future__ import annotations
 
+import json
 import logging
 import re
 from typing import Any, cast
 
-from fastapi import APIRouter, Depends, HTTPException, Response, status
+from fastapi import APIRouter, Depends, HTTPException, Query, Response, status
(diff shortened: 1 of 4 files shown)


nmgarza5:
@greptile

PR #346, opened 2026-07-06
feat(triggers): Triggers page redesign per the Figma mock

nmgarza5:
## Description

Rebuilds the Triggers page as the mock's card feed (Figma node `36:19460`). Contained tabs split update triggers from scheduled ones, a borderless search bar filters cards, and Add Trigger opens the New Trigger side panel. Each trigger is a card: scope chip, owner and destination avatar cluster, quiet info/gear controls (history modal and edit panel) with an enable Switch, and recent fire lines from the fires API. Fire messages expand per row via the chevron, rendered as inline markdown. A trigger that has never fired shows the IF/Then summary with a "No runs yet" badge. Disabled triggers render on the tint-01 shell with the switch off.

The fire view gains `destination_config_id` so each line names the exact config it delivered to (type-matching could name the wrong address when several configs share a type). The Destinations card leaves the page per the mocks; connector management lands on the Settings Connectors page in the next PR.

## How Has This Been Tested?

New field covered in `test_triggers_fires_api.py` (6 green). Verified live in Chrome against real triggers: tab switching, fire lines with correct per-config addressing, chevron expansion, disabled rendering, search, and Add Trigger opening the panel.

## Additional Options

### DIFF
--- a/backend/app/api/triggers.py
+++ b/backend/app/api/triggers.py
@@ -156,6 +156,11 @@ def _fire_view(row: Any) -> TriggerFireView:
         reason=str(payload.get("reason") or ""),
         message=str(payload.get("message") or ""),
         destination_type=str(payload.get("destination_type") or ""),
+        destination_config_id=(
+            str(payload["destination_config_id"])
+            if payload.get("destination_config_id")
+            else None
+        ),
     )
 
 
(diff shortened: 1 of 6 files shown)


nmgarza5:
@greptile

PR #348, opened 2026-07-06
feat(settings): tabbed user settings page (General / Wiki Preferences / Account & Access)

nmgarza5:
## Description

Rebuilds `/app/settings` on the SettingsLayouts shell with the left rail from the settings mock: General, Wiki Preferences, and Account & Access, with tab state in the URL (`?tab=`) so tabs deep-link. Existing settings redistribute without behavior change: theme and timezone under General, default landing and the chat model override under Wiki Preferences, display name plus the read-only login email under Account & Access.

A shared draft hook now owns the diff-against-baseline save (only changed keys PUT, preserving the unset-timezone semantics), text inputs move to Opal InputTypeIn, and the inline timezone escape link becomes LinkButton. The Connectors tab and the upcoming Notifications tab slot into this rail in their own PRs.

## How Has This Been Tested?

Verified live in Chrome: all three tabs render, rail selection tracks the URL, deep links work, Save stays disabled until a field changes. tsc and prettier clean.

## Additional Options


nmgarza5:
@greptile

PR #349, opened 2026-07-06
feat(notifications): opt-in email copies for comments and update warnings

nmgarza5:
## Description

Bo's requests: two per-user notification settings, both default off, both mailed to the login address (`users.email`) with no verification step since that address is already the account's identity.

Comment emails fan out on new comments and replies to the page owner and any @mentioned users, never the author, with mention tokens rendered as plain names and a link back to the page. Update-warning emails go to the page owner when a page crosses the auto-update warning threshold or gets paused at the cap, and they inherit those events' existing once-per-crossing and once-per-episode dedup, so a page stuck over the cap mails once, not per push.

Delivery is a queued task on the triggers worker (same trust class as the Craft launches that already run there); the opt-in check lives next to the send, so producers stay fire-and-forget and a preference flipped after enqueue is still honored. Send failures log and drop. The settings live in the existing JSONB settings blob (no migration), and the settings page gains a Notifications tab with immediate-save switches.

Stacked on #348 (needs the tab rail).

## How Has This Been Tested?

Five new tests in `test_notification_emails.py`: preference gating, unknown-kind drop, swallowed send failure, owner+mention-not-author fan-out with detokenized bodies, and the ownerless no-op. Verified live: toggling the switch PUT 200 and flipped the stored setting, then restored.

## Additional Options

### DIFF
--- a/backend/app/api/comments.py
+++ b/backend/app/api/comments.py
@@ -34,6 +34,7 @@ from app.models.comment import (
     EditCommentRequest,
 )
 from app.wiki import comment_remap, comments as comments_repo, filesystem
+from app.wiki import comment_notifications
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #355, opened 2026-07-07
feat(triggers): Activity History flyout entries per the Figma mock

nmgarza5:
## Description

Rebuilds the Activity History flyout's entries to the mock (`37:81552`). Each row now speaks the trigger cards' language: scope chip, kind glyph, and the owner/destination avatar cluster on the left, relative time with a read dot on the right (blue fill for new, hollow for older), then a collapsed "Sent a message to Slack" line with a chevron and the expandable message body. New entries open expanded by default; the frequent-update and cap events render through the same row shape. The panel chrome (header, new-count badge, search, New/Older dividers, empty states) was already Opal and stays.

The chip and avatar cluster move into a shared `fireParts` module used by both the trigger cards and the flyout, so the two surfaces stay in lockstep. One rendering bug fixed along the way, found by DOM measurement: `markdown()` inside Text's inline span leaves whitespace nodes around the block children ReactMarkdown emits, which render as phantom line boxes (a 20px line measured 68px). Message bodies now render as block paragraphs and the one-liners compose plain Texts, on both surfaces.

## How Has This Been Tested?

Verified live in Chrome against the mock: new entries expanded with tight messages and blue dots, older entries collapsed with hollow dots, chevron expands in place, search and dividers intact. tsc and prettier clean.

## Additional Options

### DIFF
--- a/frontend/src/components/triggers/TriggerCard.tsx
+++ b/frontend/src/components/triggers/TriggerCard.tsx
@@ -211,8 +211,8 @@ export function TriggerCard({
                   </div>
                 </div>
                 {isOpen && f.message && (
-                  <div className="w-full pr-5 pb-1">
-                    <Text as="p" font="main-ui-body" color="text-03">
+                  <div className="w-full pr-5 pb-1 [&>span]:block">
+                    <Text font="main-ui-body" color="text-03">
                       {markdown(f.message)}
                     </Text>
                   </div>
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #357, opened 2026-07-07
feat(settings): Connectors tab with Slack mute and the Emails modal

nmgarza5:
## Description

Adds the Connectors settings tab from the mocks (`1303:47918`, `1374:59061`, `1426:40125`, `1426:41792`), restoring the destination management that left the Triggers page, in the place the design puts it.

The Slack card shows Connect (when the admin has configured the app), or Connected with the workspace name, a mute toggle, and Disconnect. Mute pauses delivery without disconnecting: dispatch skips muted connections before any Slack I/O and still records the fire to the Activity Center, backed by a `muted` column on the connection (inspector-guarded migration) and `PUT /api/connectors/slack/mute`.

The Emails card lists address chips and opens the management modal: add-and-send, Verified / Not verified rows, a Verify resend per unverified address, and delete. Resend rate limits now return `retry_after_seconds` in the 429 body plus a Retry-After header, and the modal runs an honest per-address countdown from it. Fixing that surfaced an app-level bug: the HTTPException handler rebuilt responses without the exception's headers, so semantic headers like Retry-After never reached clients — now preserved.

## How Has This Been Tested?

Four new tests: mute round-trip, mute-without-connection 404, muted-connection dispatch skip that still records the fire, and the 429 carrying matching Retry-After header and body values. Verified live in Chrome: both cards render real connection data, the mute toggle flipped the stored flag and back, and the modal lists verified and unverified addresses. The live resend was deliberately not clicked (it would email a teammate); the countdown mechanics are pinned by the test.

## Additional Options

## Webhook mute decision table

Slack webhook URLs embed their workspace id (hooks.slack.com/services/<TEAM>/), so mute resolution is:

| Case | Behavior |
|---|---|
| URL parses to a team | Exactly that workspace's mute applies |
| URL does not parse, one connection | That connection's mute applies (it is the only Slack there is) |
| URL does not parse, multiple connections | Delivers. The webhook cannot be attributed, and borrowing another workspace's mute silences unrelated destinations |

Real Slack webhooks always parse. The unattributable case is only Slack-compatible endpoints on custom hosts, where a workspace mute has no defined relationship to the endpoint. All four cases are pinned in test_slack_mute_and_resend_retry.py.

Bot-path legacy configs (no team_id stamp) try each unmuted connection in stable order and deliver where the channel is accepted, because a wrong-workspace token fails without sending. Muted connections are never candidates, so a muted workspace cannot deliver, and an unmuted workspace is exactly where the channel lives. New configs are always stamped at creation.

### DIFF
--- a/backend/app/api/slack_connect.py
+++ b/backend/app/api/slack_connect.py
@@ -40,8 +40,14 @@ class SlackConnectStatus(BaseModel):
     configured: bool
     connected: bool
     team_name: str | None = None
+    team_id: str | None = None
     token_display: str | None = None
+    muted: bool = False
     connect_url: str | None = None
+
+
+class SlackMuteRequest(BaseModel):
+    muted: bool
 
 
 def _require_configured() -> app_settings.SlackAppSettings:
@@ -77,9 +83,25 @@ def get_status(user: User = Depends(require_user)) -> SlackConnectStatus:
         configured=settings.configured,
         connected=first is not None,
         team_name=first["team_name"] if first else None,
+        team_id=first["team_id"] if first else None,
         token_display=first["token_display"] if first else None,
+        muted=bool(first["muted"]) if first else False,
         connect_url=connect_url,
     )
+
+
+@router.put("/mute", response_model=SlackConnectStatus)
+def set_mute(
+    req: SlackMuteRequest, user: User = Depends(require_user)
+) -> SlackConnectStatus:
+    """Pause or resume Slack delivery on the user's connection without
+    disconnecting."""
+    rows = connections.list_for_user(user.id)
+    first = rows[0] if rows else None
+    if first is None:
+        raise HTTPException(status_code=404, detail="not connected")
+    connections.set_muted(user.id, str(first["team_id"]), req.muted)
+    return get_status(user)
 
 
 @router.get("/start")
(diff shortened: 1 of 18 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #361, opened 2026-07-07
Record who last modified the ingest settings / API key

Bo-Onyx:
## Summary
The ingest API key was silently rotated on dev-wiki (2026-07-07 00:23 UTC), breaking the Onyx document push hook with 401s, and the actor could not be identified — it only existed in a pod log line that died with the nightly deploy.

This adds an `updated_by_user_id` column to `ingest_settings` (FK to users, SET NULL on delete), stamps it on both the settings save and key regeneration, and shows "Last modified <time> by <email>" on /admin/ingest.

## Testing
- `pytest` on ingest/admin/craft test files (65 passed), basedpyright and ruff clean, frontend tsc clean.


Bo-Onyx:
@greptitle

PR #363, opened 2026-07-08
feat(triggers): webhook destination (dispatch, SSRF guard, HMAC signing, send-test)

nmgarza5:
## Description

The generic webhook destination from the Trigger Destinations plan (PR 9). A `type=webhook` destination config carries a URL, optional headers, a routing tag, and static fields in `config_json`, with an HMAC signing secret in the existing encrypted column. Only a catalog row is seeded, no schema change.

- On a fire, dispatch builds a structured `WebhookEvent` (trigger id/kind, doc_path, sha, change_kind, fired_at, actor, summary, reason, routing_tag, fields), HMAC-SHA256 signs the exact body, runs the URL through a new SSRF guard, and POSTs. Failures log and swallow since the fire is already recorded.
- `POST /destination-configs/{id}/test` fires a sample event so a receiver can learn the field shape.
- Create/list/delete already work through the generic `/destination-configs` CRUD.

Payload is summary + metadata only (doc body, diff, and file attachments are deferred, decision 11).

**Note:** the plan said reuse Craft's SSRF egress block, but no such guard exists in the repo, so this adds one (`app/net/ssrf.py`) rejecting private/loopback/link-local/metadata hosts. The Settings webhook card and the trigger-panel picker are separate follow-ups.

## Additional Options

### DIFF
--- a/backend/app/api/triggers.py
+++ b/backend/app/api/triggers.py
@@ -240,7 +240,7 @@ def resend_verification(
 )
 def send_test_event(config_id: str, user: User = Depends(require_user)) -> Response:
     """Fire a sample event at a webhook config so the receiver can learn the
-    payload shape. Owner-scoped; webhook configs only."""
+    payload shape. Owner-scoped, webhook configs only."""
     row = dest_configs.get(config_id, user.id)
     if row is None:
         raise HTTPException(status_code=404, detail="not found")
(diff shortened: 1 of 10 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #364, opened 2026-07-08
fix(coedit): sessions follow page moves; checkpoint never resurrects a dead path

Bo-Onyx:
## What

Closes the rename gap left after #350/#351: co-edit sessions (and their queued checkpoint tasks) are path-keyed, and nothing migrated them on a page move — so a checkpoint firing after a rename re-created the page under its old name from the session buffer. This happened three times on the Wiki Auto Management PRD (2026-07-06/08).

## How

- `coedit.on_path_moved(moves)` — re-keys `coedit_sessions.path` on every move/rename (exact pairs + folder-prefix carry), wired into `notify.after_path_move` next to the ACL/update-policy re-keying. Checkpoint tasks carry only `session_id` and resolve the path through the row, so the re-key retroactively fixes already-queued tasks too.
- Backstop in `_checkpoint_locked`: an **active** session whose seeded page (`base_sha` set) no longer exists at HEAD is closed with a warning — buffer preserved in the row for recovery — instead of committing to the dead path. `base_sha=None` (session opened on a not-yet-committed page) still creates, so the new-page flow is untouched.

Known residual: a checkpoint already mid-execution at the exact moment of a move can still write the old path (sub-second window; content is current, next checkpoint lands on the new path). Same class as other write-during-move races the merge gateway already tolerates.

## Testing

5 new tests in `tests/test_coedit_checkpoint.py`: re-key on page move, folder rename carries nested sessions, checkpoint commits to the new path after a move, dead-path session closes without resurrecting (buffer preserved), and brand-new-page creation still works. Suite: 19 passed; ruff + basedpyright clean.

### DIFF
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -375,6 +375,15 @@ def move_document_or_folder(
             status_code=409,
             detail="a file or folder with that name already exists",
         )
+    # An active co-edit session can hold a not-yet-committed draft at the
+    # destination — no file on disk, so the exists() check above misses it.
+    blocking = coedit.blocking_active_session_path(new_rel)
+    if blocking is not None:
+        raise HTTPException(
+            status_code=409,
+            detail=f"someone is editing an unsaved draft at '{blocking}' — "
+            "pick a different name or wait for it to be saved",
+        )
     if old_abs.is_file() and old_rel.endswith(".md") and not new_rel.endswith(".md"):
         raise HTTPException(
             status_code=400,
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -412,7 +412,7 @@ def move_document_or_folder(
         old_path=old_rel,
         new_path=new_rel,
         sha=sha,
-        moved=[MovedFile(old=o, new=n) for o, n in moves],
+        moved=[MovedFile(old=mv.old, new=mv.new) for mv in moves],
     )
 
 
--- a/backend/app/llm/agents/tools/move_path.py
+++ b/backend/app/llm/agents/tools/move_path.py
@@ -55,7 +55,7 @@ def handle(args: dict[str, Any]) -> Any:
             "old_path": old_rel,
             "new_path": new_rel,
             "sha": sha,
-            "moved": [{"old": o, "new": n} for o, n in moves],
+            "moved": [{"old": mv.old, "new": mv.new} for mv in moves],
         }
     except ToolError as exc:
         return {"error": str(exc)}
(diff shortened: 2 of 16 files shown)


Bo-Onyx:
Added a second commit per Bo's review: the move seam now speaks a frozen pydantic `PathMove` model (`app/models/wiki.py`) instead of raw `list[tuple[str, str]]` — producer (`git.move_path`), `common_folder_rename`, `after_path_move`, and all four re-keying consumers. Full backend suite: 1410 passed.

Bo-Onyx:
All four Greptile P1s addressed in the latest commit:

1. **Sibling re-key** — real bug, fixed by deleting the folder-prefix branch outright: sessions are keyed to `.md` files and `git.move_path` emits an exact pair per nested file, so per-pair re-keys already cover folder renames. Regression test: single cross-folder move leaves a sibling session untouched. (ACL/update-policy keep their prefix branches — they have folder-scoped rows; their single-file-move exposure is pre-existing and out of scope here.)
2. **Destination collision** — re-key now pre-checks for an active session at the destination and skips that pair with a warning instead of tripping the unique index mid-fan-out; the stranded session is retired later by the missing-path guard, buffer preserved. Test included.
3. **Transient git failure** — guard now uses a working-tree `is_file()` check (same as the read API); an OSError propagates so the task retries rather than misreading a failure as a missing page.
4. **Move race vs live session** — the guard now closes only participant-less sessions; with live participants it skips and lets the periodic scan retry after the re-key lands. Test included.

Suites: coedit checkpoint/repo/rebase + move-notify, 70 passed; ruff + basedpyright clean.

Bo-Onyx:
Follow-up from Bo's review discussion (move-vs-new-session races):

- **Prevention**: `/move` (API + agent tool) now 409s when an active co-edit session holds an unsaved draft at the destination — the file-exists check can't see uncommitted drafts (`coedit.blocking_active_session_path`, exact page or nested-under-folder).
- **Collision rule simplified per Bo: origin always wins.** With the 409 pre-check, any session still active at the destination was necessarily opened inside the seconds-wide window since validation — it can't hold much, while the origin buffer may hold a lot. The young destination session closes with its buffer preserved in the row.
- **TOCTOU hardening**: each re-key pair runs in a savepoint; a racing insert that trips the active-unique index degrades to a logged skip instead of aborting the whole move fan-out.
- Known follow-up (out of scope): a stale client can still open a session at the *old* path post-move; the checkpoint guard bounds it (skip while occupied, close when empty). Proper fix is join-time existence validation.

Suites: coedit checkpoint/repo/api + move-notify, 85 passed; ruff + basedpyright clean.

PR #365, opened 2026-07-09
feat(auth): AI system user — seeded identity principal

Bo-Onyx:
## What

The identity half of Wiki Auto Management's concept stack ([PRD — Identity and authorization](https://dev-wiki.onyx.app/app/wiki/Engineering%20Projects/Agent%20Wiki%20Project/design/PRD:%20Wiki%20Auto%20Management.md)): a seeded, first-class **AI system user** the engine will attribute commits to, receive ACL grants as, and own pages as. Identity only — nothing acts as this user yet.

- `users.kind` — `'human'` (default, all signups) | `'system'`
- Seed migration: fixed id `system-wiki-ai`, name "Agent Wiki AI", no password hash, not admin, active (`ON CONFLICT DO NOTHING`)
- Code references it via `users_repo.AI_USER_ID` / `get_ai_user()`, never email lookup

## Guards

- **Login**: cannot authenticate — no hash, plus an explicit `kind` guard in `authenticate` so even a force-set hash wouldn't help
- **First-signup auto-admin**: counts humans only, so the seeded row can't steal the promotion on a fresh install
- **Listings**: excluded from `list_all` / `search` (typeaheads) / `status_counts` (admin dashboard) by default; `list_all(include_system=True)` opts in; `get_many` still resolves it so attribution surfaces (comments, history) render the name
- **Admin API**: `PATCH`/`DELETE /admin/users/{id}` refuse system users (400), so it can't be promoted, deactivated, or deleted

## Testing

New `tests/test_system_user.py` (6 tests): seed exists post-migrations, first human still auto-admin, authenticate refused even with a forced hash, listings exclude / `include_system` opts in / `get_many` resolves, admin modify+delete refused. Full backend suite: 1436 passed (one existing test caught `status_counts` counting the AI user — fixed to count humans). ruff + basedpyright clean.

### DIFF
--- a/backend/app/api/admin.py
+++ b/backend/app/api/admin.py
@@ -121,6 +121,9 @@ def update_user(
     if req.is_active is False and user_id == actor.id:
         raise HTTPException(status_code=400, detail="cannot deactivate yourself")
 
+    if target["kind"] != "human":
+        raise HTTPException(status_code=400, detail="cannot modify a system user")
+
     # An active admin being demoted and/or deactivated must not be the last one.
     if (was_admin and was_active) and not (final_admin and final_active):
         if users_repo.admin_count() <= 1:
@@ -197,6 +200,8 @@ def delete_user(user_id: str, actor: User = Depends(require_admin)) -> OkRespons
     target = users_repo.get_by_id(user_id)
     if target is None:
         raise HTTPException(status_code=404, detail="not found")
+    if target["kind"] != "human":
+        raise HTTPException(status_code=400, detail="cannot delete a system user")
     if bool(target["is_admin"]) and users_repo.admin_count() <= 1:
         raise HTTPException(status_code=400, detail="cannot delete the last admin")
     users_repo.delete(user_id)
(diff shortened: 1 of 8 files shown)


Bo-Onyx:
Two additions from review discussion:

- **UserKind enum + DB CHECK** (earlier commit): `UserKind(str, Enum)` in `app/auth/users.py` is the single source of truth (SessionStatus/CommentStatus pattern); call sites compare against it and `users_kind_check` enforces it at the DB (model + guarded migration + rejection test). This also resolves the inline comment about missing DB-level constraint — it anchored to the pre-enum revision.
- **OIDC guard** (this commit): `upsert_oidc_user` matched any existing row by email, so an IdP asserting the AI user's placeholder email would have minted a session *as* the AI user. It now raises `SystemUserSignInError` for system principals and the callback redirects with the standard not-allowed error. Test included.

Suites: system-user + OIDC, 22 passed; ruff + basedpyright clean.

PR #367, opened 2026-07-09
feat(triggers): webhook destination UI (connectors card, manage modal, trigger picker)

nmgarza5:
## Description

Frontend for the #363 webhook destinations:

- Connectors tab gets a Webhooks card. Endpoints render as named chips through a new generic `ChipList` (same rendering as the email chips).
- `WebhookModal`: register endpoints with a name, URL, and optional routing tag. Per-endpoint Test button POSTs a sample event. Signing secret is minted server-side when the creator supplies none.
- Trigger editor: webhook destination type with an endpoint picker. Trigger cards now tag the kind (Recurring / On updates) and hide actions whose destination was deleted.
- Shared `ConnectorModal` chrome (shell, config row, mini action button, count divider) extracted and consumed by both the Emails and Webhooks modals.
- Backend: webhook delivery errors now include the receiver's response body (truncated) so a 4xx says why.

## How Has This Been Tested?

### DIFF
--- a/backend/app/triggers/destination_configs.py
+++ b/backend/app/triggers/destination_configs.py
@@ -15,6 +15,7 @@ secret is never written to the wiki git repo. Repos return dicts, not ORM rows.
 from __future__ import annotations
 
 import logging
+import secrets
 import uuid
 from datetime import datetime, timezone
 from typing import Any, cast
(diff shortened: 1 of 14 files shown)


nmgarza5:
The uncapped webhook chip list is intentional. Hiding endpoints behind a +N chip was tried and rejected during review of this feature: the card must show every registered endpoint at a glance. The chips are small and flex-wrap, so the card grows by rows rather than overflowing, and realistic endpoint counts are single digits. The email card keeps its capped treatment because that matches its mock.

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #368, opened 2026-07-09
feat(backup): opt-in scheduled backups to S3-compatible storage

Bo-Onyx:
## What

Adds an opt-in backup story for self-hosted deployments. `backup.enabled=true` in the Helm chart schedules a CronJob that uploads an **application-consistent pair** — a `git bundle` of the wiki repo and a `pg_dump` of the app database — to any S3-compatible bucket (AWS S3, MinIO, R2, GCS interop).

Both halves are required for a faithful restore: permissions, owners, update policies, and comments are Postgres-only by design, so a volume-only backup silently loses them. That trap is called out in the new `docs/backups.md`, along with the full restore procedure.

## How

- **`app/scripts/backup_to_s3.py`** — `python -m app.scripts.backup_to_s3`: git bundle + `pg_dump --format=custom`, boto3 upload under `<prefix>/<timestamp>/`, optional keep-last-N pruning. Env-configured (`BACKUP_S3_*`), creds via static keys or ambient IAM (IRSA).
- **`app/wiki/git.py`** — new `bundle()` wrapper; git stays behind the seam.
- **`backend/Dockerfile`** — `postgresql-client-17` from PGDG (bookworm ships pg_dump 15, which refuses a v17 server). No new image: the job runs on the backend image, which already has git and boto3.
- **Chart (0.3.31)** — `backup.*` values block; CronJob mounts the wiki PVC read-only and co-schedules with the backend (RWO constraint, same pattern as workers); creds land in the chart secret only when set inline; `required` guard on the bucket; NOTES.txt hint when backups are off.
- **`deploy/README.md`** — replaces the "backups not wired" Day-2 bullet.

## Testing

- `tests/test_backup_script.py`: bundle→`git clone` round-trip against a real tmp repo, upload key layout, keep-last pruning, SQLAlchemy-URL normalization for pg_dump. All pass.
- `helm lint` + `helm template` on all three config paths (enabled+creds, enabled+IRSA, required-bucket guard).
- basedpyright + ruff clean.


backend/app/scripts/backup_to_s3.py
```diff
@@ -0,0 +1,184 @@
...
+        dump_path = Path(tmp) / f"db-{ts}.dump"
+        wiki_git.bundle(str(bundle_path))
+        log.info("wiki bundle: %d bytes", bundle_path.stat().st_size)
+        dump_database(str(dump_path))
+        log.info("db dump: %d bytes", dump_path.stat().st_size)
+        keys = upload(client, cfg, ts, [bundle_path, dump_path])
+    prune(client, cfg)
+    return keys
+
+
```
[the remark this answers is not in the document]
Bo-Onyx: removed

Bo-Onyx:
Ran the full backup→restore cycle locally against the live Docker Compose stack, using MinIO as the S3-compatible target (i.e. the exact customer path — image built from this branch, read-only wiki mount, env-only config, no AWS):

- **Backup**: git bundle (96 KB) + pg_dump (157 KB) uploaded under one timestamp group; read-only wiki mount works; PGDG pg_dump 17 talks to Postgres 17.
- **Restore**: `git clone` of the bundle reproduced the committed wiki tree exactly (only an untracked dir differed, as expected for a git backup); `pg_restore` into a fresh DB matched the original row-for-row across all 48 tables (users/acl/comments/triggers/chat/update_policies spot-checked).
- **Pruning**: three runs with `keepLast=2` left exactly the two newest groups.

The run also caught a real bug, fixed in 6d7035f: importing `app.config` validates all required fields (`PUBLIC_BASE_URL` among them), so the CronJob crashed with only `DATABASE_URL`/`WIKI_DIR` set. The template now reuses the `backendEnv` helper so the job boots with the same env as the backend — robust against future required fields too.

backend/app/scripts/backup_to_s3.py
```diff
@@ -0,0 +1,220 @@
...
+    url = CONFIG.database_url.replace("postgresql+psycopg://", "postgresql://", 1)
+    parts = urlsplit(url)
+    env = None
+    if parts.password is not None:
+        netloc = parts.username or ""
+        netloc += f"@{parts.hostname}" if parts.hostname else "@"
+        if parts.port:
+            netloc += f":{parts.port}"
+        url = urlunsplit(parts._replace(netloc=netloc))
+        env = {**os.environ, "PGPASSWORD": parts.password}
```
nmgarza5: `urlsplit().password` is still percent-encoded and PGPASSWORD is taken literally, so any DB password that needed encoding in DATABASE_URL (`@`, `%`, `:`) fails auth on every backup while the app itself works fine. The netloc rebuild also drops IPv6 brackets. Fix: parse with `sqlalchemy.engine.make_url` (already a dependency), export the decoded `url.password`, and render the URL with the password removed. That also covers `+psycopg2`/`+asyncpg` dialect markers instead of the one hardcoded `.replace`.

backend/app/scripts/backup_to_s3.py
```diff
@@ -0,0 +1,220 @@
...
+        log.info("pruned %d object(s) beyond keep_last=%d", len(deleted), cfg.keep_last)
+    return deleted
+
+
+def run_backup(cfg: BackupConfig, client: Any) -> list[str]:
+    ts = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+    with tempfile.TemporaryDirectory(prefix="agent-wiki-backup-") as tmp:
+        bundle_path = Path(tmp) / f"wiki-{ts}.bundle"
+        dump_path = Path(tmp) / f"db-{ts}.dump"
+        wiki_git.bundle(str(bundle_path))
```
nmgarza5, line 144: The bundle-before-dump order is load-bearing: the dump can only be newer than the bundle, so a page created in the gap restores as orphan ACL rows. Flipped, it would restore as a page with no ACL rows, and the implicit-public fallback in `app/wiki/acl.py` makes that page public. Fix: a comment here pinning the order.

backend/app/scripts/backup_to_s3.py
```diff
@@ -0,0 +1,220 @@
...
+        env=env,
+    )
+
+
+def upload(client: Any, cfg: BackupConfig, ts: str, paths: list[Path]) -> list[str]:
+    """Upload each file to ``s3://bucket/prefix/ts/<name>``; return the keys."""
+    keys: list[str] = []
+    for p in paths:
+        key = f"{cfg.prefix}/{ts}/{p.name}"
+        client.upload_file(str(p), cfg.bucket, key)
```
nmgarza5, line 127: If the dump upload fails after the bundle uploaded, the half group sits in the bucket forever, counts toward `keep_last`, and looks restorable. Fix: have `prune` only count groups containing both files and clean up incomplete ones, or upload a small manifest last and treat only manifested groups as backups.

backend/app/scripts/backup_to_s3.py
```diff
@@ -0,0 +1,220 @@
...
+        page = client.list_objects_v2(
+            **kwargs, **({"ContinuationToken": token} if token else {})
+        )
+        yield page
+        token = page.get("NextContinuationToken")
+        if not token:
+            return
+
+
+def prune(client: Any, cfg: BackupConfig) -> list[str]:
```
nmgarza5: Consider dropping prune in favor of the bucket lifecycle rule the docstring already recommends: that deletes `_list_all` plus the FakeS3 pagination machinery, and takes DeleteObject out of the IAM grant, which matters given the shared ServiceAccount. If it stays: boto3's `get_paginator("list_objects_v2")` replaces `_list_all`, `delete_objects` batches 1000 keys per call, and any foreign sub-prefix under the prefix currently counts as a backup group (a `manual/` folder would permanently occupy a keep_last slot while a real backup gets deleted).

deploy/helm/agent-workspace/templates/backup-cronjob.yaml
```diff
@@ -0,0 +1,99 @@
...
+      backoffLimit: 2
+      activeDeadlineSeconds: 3600
+      template:
+        metadata:
+          labels:
+            {{- include "agent-workspace.labels" . | nindent 12 }}
+            app.kubernetes.io/component: backup
+        spec:
+          restartPolicy: Never
+          serviceAccountName: {{ include "agent-workspace.serviceAccountName" . }}
```
nmgarza5: This is the same ServiceAccount the backend and workers run on, so the IRSA setup documented in docs/backups.md grants PutObject/ListBucket/DeleteObject on the backup bucket to every pod running agent-driven code. A compromised backend can delete the exact backups that protect against it. Fix: a dedicated ServiceAccount for the CronJob with its own annotations value.

deploy/helm/agent-workspace/templates/backup-cronjob.yaml
```diff
@@ -0,0 +1,99 @@
...
+    spec:
+      backoffLimit: 2
+      activeDeadlineSeconds: 3600
+      template:
+        metadata:
+          labels:
+            {{- include "agent-workspace.labels" . | nindent 12 }}
+            app.kubernetes.io/component: backup
+        spec:
+          restartPolicy: Never
```
nmgarza5, line 24: backend-deployment and worker-deployment both render `.Values.podSecurityContext` and `.Values.securityContext`, this pod template renders neither. Anyone setting fsGroup/runAsUser to make the wiki PVC readable gets a backup pod that can't read /wiki, and restricted Pod Security admission rejects it outright. Fix: include both blocks like the sibling templates.

deploy/helm/agent-workspace/templates/backup-cronjob.yaml
```diff
@@ -0,0 +1,99 @@
...
+    app.kubernetes.io/component: backup
+spec:
+  schedule: {{ .Values.backup.schedule | quote }}
+  concurrencyPolicy: Forbid
+  successfulJobsHistoryLimit: 3
+  failedJobsHistoryLimit: 3
+  jobTemplate:
+    spec:
+      backoffLimit: 2
+      activeDeadlineSeconds: 3600
```
nmgarza5: The bundle and dump are staged on container ephemeral storage with no ephemeral-storage request, so a large wiki risks kubelet eviction on the very node the backend is pinned to, and this 3600s deadline has no values knob. Fix: add ephemeral-storage to `backup.resources` and expose the deadline in values.

docs/backups.md
```diff
@@ -0,0 +1,100 @@
...
+
+To take a one-off backup outside the schedule:
+
+```bash
+kubectl create job --from=cronjob/<release>-agent-workspace-backup manual-backup-1
+```
+
+## Restore
+
+Restore both halves **from the same timestamp group**, then let the app
```
nmgarza5: The restore is missing the encryption key: SECRET_KEY (or ENCRYPTION_KEY_SECRET) derives the AES key for encrypted DB columns like LLM provider credentials, and it lives in the Helm secret, not in this backup. A rebuild-from-scratch restore with a freshly generated key succeeds with every encrypted column unreadable. Fix: a step requiring the original Helm secret values to be preserved with the backups.

docs/backups.md
```diff
@@ -0,0 +1,100 @@
...
+    secretAccessKey: "..."
+    # ... or leave both empty and bind an IAM identity to the chart's
+    # service account (e.g. EKS IRSA) with PutObject/ListBucket/DeleteObject
+    # on the bucket.
+```
+
+Each run writes `s3://<bucket>/<prefix>/<timestamp>/wiki-<timestamp>.bundle`
+and `db-<timestamp>.dump`, then (with `keepLast` > 0) prunes all but the
+newest N timestamp groups. Failures surface as failed Jobs — alert on
+`kube_job_status_failed` for the `<release>-backup` CronJob, or check
```
nmgarza5: `kube_job_status_failed` misses backups that stop being scheduled at all (suspended CronJob, `backup.enabled` flipped off on an upgrade). Fix: also recommend a freshness signal, `kube_cronjob_status_last_successful_time` or the age of the newest object under the prefix.

nmgarza5 (approved):
a few comments but the approach seems good to me

Bo-Onyx:
@nmgarza5 all nine addressed in fc0bd01 — thanks, the percent-encoded-password one was a real auth-breaker and the SECRET_KEY restore gap was the kind that only hurts years later.

- **make_url**: `dump_database` now parses with `sqlalchemy.engine.make_url` — decoded password into `PGPASSWORD`, IPv6 brackets survive, any dialect marker normalized. Tests cover an encoded password (`s3%40cr%3At` → `s3@cr:t`) and an `[::1]` host.
- **Prune dropped** in favor of bucket lifecycle rules, as you suggested: `keepLast`, `_list_all`, and the FakeS3 pagination are gone; the job only ever writes, and DeleteObject is out of the IAM story. Half-uploaded groups can't displace complete backups anymore; docs state that a group missing `db-*.dump` is a failed run, not a backup (bundle uploads first, order asserted in tests).
- **Ordering comment** added in `run_backup` — dump-newer-than-bundle → orphan ACL rows, flipped → implicit-public pages.
- **Dedicated ServiceAccount** `<fullname>-backup` with its own `backup.serviceAccount.annotations` for IRSA; app pods no longer share bucket credentials.
- **Pod hardening**: `podSecurityContext`/`securityContext` rendered like the sibling templates, `ephemeral-storage` requests/limits (1Gi/4Gi defaults), `activeDeadlineSeconds` exposed in values.
- **Docs**: restore prerequisite on preserving `SECRET_KEY`/`ENCRYPTION_KEY_SECRET` (encrypted columns unreadable under a fresh key), and freshness alerting (`kube_cronjob_status_last_successful_time` / newest-object age) alongside job-failure alerting.

Re-verified the real path end-to-end after the refactor (image from branch + MinIO + live compose stack): bundle + dump uploaded clean.

PR #371, opened 2026-07-10
feat(triggers): Craft destination — a trigger fire starts an Onyx Craft session

nmgarza5:
## Description

PR 11 of the trigger-destinations plan, the last unshipped item.

- New `craft` destination type: dispatch starts an Onyx Craft session for the trigger's owner, seeded with the rendered summary + fire context, with the source page attached. Reuses the existing connection, idempotent in-flight probe, provisioning cap, and launch task.
- The launch workflow moved out of the `post_launch` route into `app/launchers/craft.py` (typed precondition errors) so the route and the dispatcher share one implementation. Route behavior and status codes unchanged.
- Craft configs are one-per-user with no secret (the Onyx connection carries the target). Catalog row seeded by migration.
- UI: Onyx Craft card on Settings → Connectors (mirrors the Slack card), Onyx Craft option in Then Send gated on the connection, selection wires the single config automatically.
- Launch failures are logged and swallowed at dispatch: the fire is always recorded first.

## How Has This Been Tested?

PR #373, opened 2026-07-10
fix(db): re-parent the template migration after the craft head

nmgarza5:
## Description

#371 and #372 both chained migrations from `c7d3e85f1b02`, leaving two alembic heads, so `init_db`'s `upgrade head` fails at boot (dev-wiki backend crash-loops with "Multiple head revisions").

Fix: re-parent `e4b8c61d9a73` (template ai management) onto `d4f8a26e9c11` (craft destination) for a single linear chain. Safe because no environment is stamped on a divergent head: dev-wiki's `alembic_version` reads `c7d3e85f1b02` (checked on the live DB), so a redeploy applies both migrations in order.

Validated: single `alembic heads`, fresh per-test schemas build the linear chain.

Redeploy after merge to heal dev-wiki.

## How Has This Been Tested?


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #374, opened 2026-07-10
feat(wiki): tombstone lookup + restore for deleted pages

Bo-Onyx:
## Problem

When a page or folder is deleted (or moved), its old URL becomes a bare "not found" error with no in-product recovery. The content is never actually lost — `delete_path` is an additive git commit — but nothing lets a user discover the delete commit, view the old content, or bring the page back; recovery requires an operator with shell access to the wiki volume.

## What this adds (backend only)

- **`path_fate()`** (`app/wiki/git.py`): resolves a HEAD-absent path to its fate — `deleted` (delete commit + the parent ref where content is still readable) or `moved` (rename chains followed forward to the current name; moved-then-moved and moved-then-deleted chains resolve to their final fate). Works for files and folders; a folder's fate is derived from its tracked children.
- **`restore_path()`** (`app/wiki/git.py`): `git checkout <ref> -- <path>` + commit under the commit lock — additive history, no revert.
- **`GET /api/wiki/file/tombstone?path=…`**: fate of a missing path, incl. `last_content_sha` (readable via existing `GET /wiki/file?ref=`) and `can_restore`.
- **`POST /api/wiki/file/restore`**: restores a deleted file or folder as of just before its delete commit. Each restored `.md` runs the normal create lifecycle (`after_doc_write(change_kind=create)`) so FTS, triggers, and MCP pub-sub fire and the restorer becomes the owner with fresh default ACLs. Restored `.trigger_*.yaml` files reconverge the trigger cache.

## Permissions stance

Deleting a page hard-deletes its ACL rows, so tombstone reads resolve through the closest surviving ancestor folder's ACLs (implicit-public when unconfigured) — same resolution `require_can` already applies to unmanaged paths. For a moved path the target's ACLs gate the redirect info. Restore requires `write` on every page being restored. Pre-delete page-level grants are not resurrected (content-only restore; soft-deleting ACL/policy rows is a tracked follow-up).

## Not in this PR

- Frontend tombstone 404 view / restore button / move redirect — stacked PR next.
- Design doc: wiki page `Engineering Projects/Agent Wiki Project/design/Deleted Page Recovery.md`.

## Testing

13 new tests in `tests/test_tombstone_restore.py` covering the fate resolver (file/folder delete, move, move→delete and move→move chains), the tombstone endpoint (deleted/moved/404/409), and restore roundtrips (content + owner + additive history, folder restore, restore-at-final-name, moved-path refusal). Neighboring suites (`test_delete_notify`, `test_move_notify`, `test_acl`, `test_wiki_filesystem`, `test_wiki_diff_api`) pass; ruff + basedpyright clean.

### DIFF
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -42,6 +42,7 @@ from app.models.file_system import (
     RecordRecentDocRequest,
     ReindexRequest,
     ReindexResponse,
+    ResolveDocIdResponse,
     ReorderStarredRequest,
     ReviseDraftRequest,
     ReviseDraftResponse,
@@ -58,6 +59,7 @@ from app.wiki import (
     agent_activity,
     coedit,
     diff as wiki_diff,
+    doc_ids,
     drafts as wiki_drafts,
     filesystem,
     git as wiki_git,
@@ -71,7 +73,7 @@ from app.wiki import (
 )
 from app.ingest import settings as ingest_settings
 from app.models.update_policy import UpdateHealthResponse
-from app.models.wiki import ChangeKind, CommitMaxRetriesError
+from app.models.wiki import ChangeKind, CommitMaxRetriesError, PathMove
 
 router = APIRouter()
 log = logging.getLogger(__name__)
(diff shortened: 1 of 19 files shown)


Bo-Onyx:
Superseded — closing. Trash + lossless restore shipped in #382; the stable-doc-id / id-URL work is re-landed cleanly on `main` in #386. This branch (`bo/deleted-page-recovery`) is retained for reference but won't be merged (it conflicts with what's since landed — trash, PageKind, the alembic head). See #386.

PR #381, opened 2026-07-13
feat(wiki): deleted-page recovery + Trash (git-based)

Bo-Onyx:
Re-lands the deleted-page **restore engine** (salvaged from the abandoned #374, stripped of the id-URL/doc_ids coupling that conflicted with main) as a clean, focused PR on `main`, plus a new **Trash** listing.

## Why git-based

Every wiki delete is an additive git commit, so deleted content is already recoverable from history — no soft-delete table needed. Restore re-commits the old content; nothing is ever force-removed.

## Backend

- **`git.py`**
  - `path_fate(path)` — for a path absent at HEAD, resolves `deleted` (with `last_content_sha`, the ref where the pre-delete body is readable) or `moved` (following rename chains forward). Files and folders.
  - `restore_path(path, ref)` — `git checkout <ref> -- <path>` + commit; additive, no revert.
  - `recent_deletions(limit)` — walks `git log --diff-filter=D` for delete events.
- **Endpoints**
  - `GET /api/wiki/file/tombstone?path=` — a missing path's fate (deleted vs moved).
  - `POST /api/wiki/file/restore` — restore a deleted file/folder; each restored `.md` runs the normal create lifecycle (`after_doc_write(create)` → default ACLs + owner stamp, FTS reindex, trigger fan-out), and restored trigger YAMLs reconverge the cache.
  - `GET /api/wiki/trash` — deleted pages not since re-created or restored, ACL-filtered (an item shows only if the caller can read its closest surviving ancestor scope).

## Scope / decisions

- **Restore is content-only** — permissions re-seed to defaults (lossless permission-restore would need soft-deleted ACL rows; deferred).
- **Trash lists pages**; a deleted folder surfaces as its nested pages for now (folder-level grouping is a follow-up).
- **Frontend Trash view** lands in a stacked PR.
- Deliberately **excludes** the stable-doc-id / id-URL feature (separate, de-prioritized).

## Testing

Ported #374's tombstone/restore tests (git-only, no doc_ids) + added Trash-list tests (deleted pages listed; re-created/restored paths excluded). ruff + basedpyright clean locally. **pytest runs in CI** (my local Docker Postgres is wedged this session, so CI is the authoritative test run).

### DIFF
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -32,6 +32,7 @@ from app.models.file_system import (
     MovedFile,
     MovePathRequest,
     MovePathResponse,
+    PathTombstoneResponse,
     PutDocumentRequest,
     PutDocumentResponse,
     RecentDocsResponse,
@@ -39,6 +40,8 @@ from app.models.file_system import (
     ReindexRequest,
     ReindexResponse,
     ReorderStarredRequest,
+    RestorePathRequest,
+    RestorePathResponse,
     ReviseDraftRequest,
     ReviseDraftResponse,
     SearchHitView,
@@ -46,6 +49,9 @@ from app.models.file_system import (
     SetDocumentDraftRequest,
     StarDocRequest,
     StarredDocsResponse,
+    TombstoneCommit,
+    TrashItem,
+    TrashListResponse,
 )
 from app.tasks.reindex import index_path
 from app.triggers import repo as triggers_repo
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
Superseded by #382 (Trash as move-to-`.trash/`, which restores losslessly — permissions and all — vs this PR's content-only, git-log approach). Closing unmerged; `path_fate` here can be salvaged later for a deleted-URL tombstone view if we build one.

PR #405, opened 2026-07-15
feat(model-server): FastAPI /score service + Dockerfile + CI (B.2b)

Bo-Onyx:
## What

**Phase B.2b** — turns the scoring core (#403) into a **running service**. Builds on the `TwoTowerScorer` HTTP client (#402): this is the server that client talks to.

## Pieces

- **`server.py`** — FastAPI app. Loads the bundle **once at startup** from `MODEL_BUNDLE_PATH` into a `BundleScorer` (lifespan), then:
  - `POST /score` — `{doc_vec, page_vecs}` → `{probs}` (mirrors the backend's wire contract)
  - `GET /health` — `{ready: bool}`
  - Startup **fails fast** if `MODEL_BUNDLE_PATH` is unset; `/score` returns 503 if the model isn't loaded.
- **`Dockerfile`** — torch (PyTorch) lives *only* in this image; deps installed from `pyproject.toml` via `uv`, uvicorn CMD on port 9100. The trained bundle is **not baked in** — supplied at runtime at `MODEL_BUNDLE_PATH` (mount / artifact download; B.3).
- **`.github/workflows/model-server.yml`** — dedicated CI (ruff + basedpyright + pytest, **with torch**) gated on `model_server/**`. The model server had *no* CI before, since the backend hooks are scoped `^backend/`.
- **Tests** — `conftest.py` builds synthetic bundles (shared by scorer + server tests). `test_server.py` exercises `/score` + `/health` via `TestClient`, and the missing-`MODEL_BUNDLE_PATH` startup failure. `test_scorer.py` refactored onto the shared fixtures (removes the duplicated builders).

## Contract note

`ScoreRequest`/`ScoreResponse` are defined here *and* in the backend (`two_tower_scorer.py`) — each side owns its DTO across the service boundary; they agree on the JSON shape. That's normal for a microservice seam (the backend can't import this torch project, and vice-versa).

## Not in scope

- **B.3** — weights delivery (S3/mount + init container) + Helm opt-in wiring + `RELEVANCE_MODEL_SERVER_URL` config on the backend. A real Docker image build also lands there.
- Wiring the filter into the reconcile flow.

## Verification

Local: **12 tests pass** (10 scorer + 2 server), ruff + strict basedpyright clean, real bundle loads + scores. The **Dockerfile is inspection-verified only** — a torch image build is heavy and there's no daemon here; the actual build gets validated with the deployment wiring (B.3). The uv flags used (`--no-install-project`, `--no-sync`) were confirmed valid.

### DIFF
--- a/.github/workflows/model-server.yml
+++ b/.github/workflows/model-server.yml
@@ -1,0 +1,37 @@
+name: Model server checks
+
+concurrency:
+  group: model-server-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+on:
+  pull_request:
+    # Self-contained project with its own (torch) deps — only run when it changes.
+    paths:
+      - "model_server/**"
+  workflow_dispatch:
+
+jobs:
+  checks:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    defaults:
+      run:
+        working-directory: model_server
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+
+      - name: Install deps
+        run: uv sync --group dev
+
+      - name: ruff
+        run: uv run --group dev ruff check two_tower tests server.py
+
+      - name: basedpyright
+        run: uv run --group dev basedpyright two_tower tests server.py
+
+      - name: pytest
+        run: uv run --group dev pytest -q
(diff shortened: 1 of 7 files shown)


Bo-Onyx:
Closing — pivoting to **in-process ONNX** instead of a separate model-server pod. The model is a small (~6M-param) CPU MLP that's internal/opt-in, so a dedicated PyTorch service pod (image + weights delivery + Deployment/Service) is disproportionate; Onyx separates its model server because its models are GB-scale/GPU/shared, which doesn't apply here. Instead: export the trained bundle torch→ONNX and run it with onnxruntime (~40MB) inside the documents worker — no new pod, same scores, equal-or-better latency. The Scorer boundary (#397) makes this a clean swap. model_server becomes the offline torch→ONNX export tool; a new OnnxScorer lands in the backend.

PR #419, opened 2026-07-16
feat(ingest): shadow-mode relevance filter in reconcile

Bo-Onyx:
## What

Wires `build_relevance_filter()` into the ingest reconcile as a **shadow pass** — the first time the relevance filter is actually consulted on real traffic. It scores what the embedding filter *would* drop, records it, and **changes nothing**: the reconciler still processes every candidate.

## Where

In `_reconcile_pushed_document`, right after the candidate set is finalized (post source-filter → BM25 → policy/cap) and **before** the LLM stages (selector + batch reconcile). That's exactly where the filter would run in enforce mode — it's cheaper than the LLM selector — so the shadow drop-set is measured against the real pre-LLM candidate set.

```
BM25 candidates → policy/cap → [SHADOW relevance filter] → LLM selector → LLM reconcile
                                        observes, drops nothing
```

## Behavior

- **Per-page metric** `ingest_relevance_shadow_total{decision=kept|dropped, wiki_path}` — drop-rate visible in Grafana.
- **Would-be drops** written as `ingest_eval_samples` rows tagged `filtered_by_relevance` (new `Text` outcome value — no migration; the column is free text). Sits alongside the existing `filtered_by_selector` / `filtered_by_bm25_score` samples for side-by-side eval review.
- **Gated on `ingest_eval_logging`** — the pass costs one document-embed call, so only data-collecting deployments run it (dev-wiki already has it on). Page vectors come from the existing `page_embeddings` store (no per-page API calls).
- **Fail-open, best-effort** — any failure (no embedding, scorer error, DB hiccup) is swallowed; shadow observation never perturbs a live ingest. Matches the filter's own fail-open contract.

## What it does *not* do

- It never drops a candidate. Enforcement is a separate, gated follow-up.
- With no model configured, `build_relevance_filter()` returns the cosine cold-start filter — so shadow data starts flowing immediately, before any `.onnx` is deployed.

## Verification

- New `test_relevance_shadow.py`: records would-be drops only (metric + `filtered_by_relevance` sample for the dropped path, nothing for the kept one); fails open (no raise, no rows) when the filter build throws.
- 59 existing ingest tests (pipeline, selector, factory) still green — no behavior change to the reconciler. Ruff + strict basedpyright clean.

## Next

- **Enforce mode** — a gated switch to actually drop the shadow-identified candidates, once the shadow data validates the operating point.
- Pairs with the delivery (#412, merged) and export tool (#408) for the warm two-tower model; cosine shadow needs neither.

### DIFF
--- a/backend/app/ingest/eval_sample.py
+++ b/backend/app/ingest/eval_sample.py
@@ -46,6 +46,10 @@ def log_sample(
         "filtered_by_selector",
         "filtered_by_bm25_score",
         "filtered_by_search_rank",
+        # Shadow only: the relevance filter would have dropped this pair before
+        # the LLM stages. The reconciler still processed it (shadow changes
+        # nothing) — so this label marks would-be drops, not actual ones.
+        "filtered_by_relevance",
     ],
     bm25_score: float | None,
     commit_sha: str | None,
--- a/backend/app/metrics.py
+++ b/backend/app/metrics.py
@@ -69,6 +69,15 @@ ingest_outcomes_total = Counter(
     ["outcome", "wiki_path"],
 )
 
+ingest_relevance_shadow_total = Counter(
+    "ingest_relevance_shadow_total",
+    "Shadow-mode relevance-filter decisions, one per candidate page. The filter "
+    "is evaluated before the LLM stages but never actually drops anything — "
+    "'dropped' counts pages it *would* filter out, 'kept' the rest. Lets us "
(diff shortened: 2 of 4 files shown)


Bo-Onyx:
Superseded: the shadow-over-BM25 approach was replaced by the service split (#430) + direct candidate-stage cutover (see the new PR). The filter now runs live as the candidate stage rather than shadowing BM25's.

PR #431, opened 2026-07-16
Always-on WYSIWYG editor with autosave

raunakab:
## Summary
- Removes the read/edit toggle in `FileView` — the WYSIWYG `Coeditor` is now always mounted for the current doc version (`DiffView` still owns history), with autosave (idle + hard-cap timers) replacing the explicit Save/Done button.
- Renames `lib/coeditor/` to `lib/editor/`, which now owns the whole editing engine: collab protocol, WYSIWYG decorations, peer presence, and comment anchoring. Comment selection/highlighting/scroll-to is ported from DOM/react-markdown offset alignment to CodeMirror-native offsets and decorations (`lib/editor/comments.ts`), replacing `commentAnchor.ts`/`rehypeSourcePos.ts`.
- Renaming moves to `DocTitle` via Opal's `Content editable`/`onTitleChange`, decoupled from the coedit session.
- Dissolves `lib/fileview/`: `FileView` (a full page view with a single consumer) moves to `views/wiki/FileView.tsx` alongside `WikiPage.tsx`; the remaining hooks/types/utils merge into `lib/wiki.ts`.
- Splits `lib/wiki.ts` into `lib/wiki/{types,constants,utils,svc,hooks}.ts`, matching `lib/editor/`'s file-per-kind convention, and updates every import site.


raunakab:
Addressed both in 120d8ba:

- **Silent join failure**: `useCoeditSession` now distinguishes a join-handshake failure (sets `joinError`, session never opens) from a stream drop after a successful join (silently ignored — the session/doc are already usable). `FileView` shows the error + a Retry button instead of an indefinite "Connecting…".
- **`leave()`/`onEnd` dead code**: folded `onEnd` into `stop()` (the only teardown path, driven entirely by the join effect's cleanup), so it now fires on every real session end. Dropped the unused `leave()` from the hook's public API, and the now-write-only `joinPromise` ref that only `leave()` read.

PR #450, opened 2026-07-17
fix(wiki): breadcrumb and blank page card fixes

nmgarza5:
## Description

Merges into main. Two fixes:

**Breadcrumb fold** (`src/chromes/header/WikiHeader.tsx`) — the crumb trail never wraps. Paths deeper than five crumbs collapse their middle ancestors behind a dots chip whose fixed 240px popover lists the hidden folders root to leaf, one line each with an ellipsis, and a hover tooltip shows the full name only for truncated rows. Includes opal 0.1.17 shims (wrapper-div ref, title-row min-width, native `title` attr strip), each verified by rendering the real installed components in a browser harness and marked for deletion when the upstream Opal fixes land (onyx #13153).

**Blank page card** (`StartNewPage`) — matches mock 709:259996: filled select-card state, no border, radius 12, chip shadow on the grid cell, title at `text-03`. The glyph renders in Opal's `IconContainer`, layering the published square and plus icons until Opal ships `SvgSquarePlus`.

No new icon files anywhere. The icons are added to Opal itself (onyx #13153). The Home crumb stays text until the `home` glyph ships.

## Additional Options

- [ ] This change should be considered for cherry-picking to the release branch
- [x] Override Linear Check

PR #440, opened 2026-07-17
feat(wiki): doc side panel refactor

nmgarza5:
## Description

Doc-page side surfaces rebuilt to the Figma Side Panel designs (section 1899:307841). One commit per surface:

1. **Tabbed side panel** — one 360px right-rail host with Updates / Comments / Sources / Watching tabs replaces the separate drawers. Sources and Watching are placeholders until their backends exist.
2. **Bottom-center chat bar** — the corner FAB and floating bubble become the mock's pill: Chat/Edit/Automations segments, dock and collapse, read-only model chip, composer. Collapsed it is the three-icon mini pill.
3. **Docked chat panel** — segmented header, tint user pills, chromeless assistant markdown, tool calls collapse into an expandable step count, copy action per reply.
4. **Doc header** — presence avatars (co-edit sessions) and an agent status avatar whose ring tracks auto-edit health, one side-panel toggle. Also fixes stale co-edit participants lingering after a session ends.
5. **Auto-edit limits** — two-card Updates tab (policy toggles + Update History with a usage meter and state chrome) and the Auto-Edit Limit modal for the per-page alert threshold.
6. **Comments panel** — search, avatar rows, foldable Resolved section, comment-count end marker.
7. **Version history** — the version list moves into the Updates tab, inline doc diff unchanged.

Disabled placeholders where no backend exists yet (Edit/Automations chat segments, attach, model switching, Organize toggle, feedback/regenerate, unread dots). Watching stays a stub for the same reason.

## Additional Options

- [ ] This change should be considered for cherry-picking to the release branch
- [x] Override Linear Check

PR #452, opened 2026-07-18
fix(wiki): fold popover truncation and tooltip polish

nmgarza5:
## Description

Breadcrumb fixes split out of #450 when it retargeted to main. These depend on `WikiHeader.tsx`, which is created by the #439/#440 stack, so this PR stays stacked on `nikg/wiki-side-panel-tabs` and merges after #440.

- **Fold popover** — fixed 240px width, rows clamp to one line via `titleMaxLines`, hover tooltip with the full name only on truncated rows.
- **Opal 0.1.17 shim** — verified by rendering the real installed components in a browser harness: the ref moves to a wrapper div (LineItemButton's ref prop reaches no DOM node), the title row's min-width is freed so the ellipsis lands inside the popover clip, and the native `title` attribute is stripped so the Opal tooltip is the only hover surface. Upstream fixes in onyx #13153.

No new icon files. The home-icon crumb waits for Opal to ship the `home` glyph (onyx #13153), and the Blank page card commit carried here (merging via #450, patch-identical, auto-drops on the post-merge rebase) layers the published square and plus icons inside `IconContainer` until `SvgSquarePlus` ships.

## Additional Options

- [ ] This change should be considered for cherry-picking to the release branch
- [x] Override Linear Check

### DIFF
--- a/frontend/src/components/wiki/StartNewPage.module.css
+++ b/frontend/src/components/wiki/StartNewPage.module.css
@@ -21,8 +21,16 @@
   gap: 8px;
   padding: 8px;
 }
+/* SelectCard strips className and style, so the mock's chip shadow and a
+   radius matching the card's corners live on the grid cell instead. */
 .templateCell {
   display: flex;
+  min-width: 0;
+  border-radius: var(--radius-12);
+  box-shadow: var(--shadow-chip);
+}
+.templateCell > * {
+  flex: 1;
   min-width: 0;
 }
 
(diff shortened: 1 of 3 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
Superseded: all of these fixes now live in nikg/wiki-ui-fixes (#450) against main's header at src/chromes/header/WikiHeader.tsx. The stacked copy stays on the branch for #440-based testing and reconciles at rebase time.

PR #459, opened 2026-07-20
fix(editor): always hide markdown markup, never reveal on caret or focus

raunakab:
## Summary
- Removes the Typora-style "reveal raw markdown while the caret/selection touches it" behavior from `wysiwygMarkdown()` (`frontend/src/lib/editor/wysiwyg.ts`) — markup (`#`, `**`, `` ` ``, `> `, list markers, task markers, `---`) now stays hidden/rendered unconditionally, regardless of caret position or editor focus.
- Deletes `spanRevealed`/`lineRevealed`/`revealScope` and the `view.hasFocus`/`update.focusChanged` gating that #458 had added on top of the same reveal mechanism — there's no reveal path left at all now.
- Rebuilt on top of latest `main` (post-#451/#454/#455/#456/#458), so it also drops the newer `TaskMarker` case's reveal check.

### DIFF
--- a/frontend/src/lib/editor/wysiwyg.ts
+++ b/frontend/src/lib/editor/wysiwyg.ts
@@ -1,11 +1,11 @@
-/** Typora-style live-preview WYSIWYG decorations for the markdown editor.
+/** Live-preview WYSIWYG decorations for the markdown editor.
  *
  * Walks the `@lezer/markdown` syntax tree on every doc/selection change and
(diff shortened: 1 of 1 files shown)


raunakab:
Good catch — fixed in 3a9c541. `update.selectionSet` was leftover from before the always-hidden change; `buildDecorations` doesn't read `state.selection` anymore so it was rebuilding decorations on every cursor move for no reason. Applied your suggested condition (`docChanged || viewportChanged`).

PR #446, opened 2026-07-17
feat(wiki): source ranges, map ingested content to its source

nmgarza5:
## Description

Stacked on #422 (review that first; this PR's diff is scoped to source ranges).

The ledger says which document an ingest commit came from, but not which part of the page it touched. This adds `source_ranges`: one row per span an ingest commit changed, linked to its ledger row and anchored like a comment (`doc_path` + `anchor_sha` + offsets). Captured at the commit gateway by diffing the parent body against the result; `record()` now returns the ledger id so ranges attach to it.

A shared `anchor_remap` skeleton (extracted from `comment_remap`, which now uses it too) re-anchors ranges on every later commit and retires one whose span is rewritten. Page moves re-point ranges alongside the ledger.

Nothing reads `source_ranges` yet. The Sources tab and highlight consume it in the frontend phase.

Design: `Engineering Projects/Agent Wiki Project/design/source_attribution/source_attribution_plan`

## How Has This Been Tested?

- `pre-commit run --files` on all changed files: ruff + basedpyright (strict) passed
- `basedpyright` project-wide: 0 errors, 0 warnings
- `pytest -xv tests/test_source_ranges.py`: 12 passed (incl. remap advance + retire over a tmp git repo)
- Regression: comment_remap, comment_notify, provenance, gateway/ingest/move/trash all green

## Additional Options

- [ ] [Optional] Please cherry-pick this PR to the latest release version.
- [x] [Optional] Override Linear Check

PR #464, opened 2026-07-20
fix(db): resolve forked alembic heads (chain source_ranges after coedit)

nmgarza5:
## Description

Main has two alembic heads, so `alembic upgrade head` fails. That runs in `init_db` on every boot and in CI, so this blocks deploys and every PR's tests.

- **Bug:** #460 (`b3e8f5a90c27` coedit) and #446 (`d8b2f1a05c93` source_ranges) both set `down_revision = c7a1f0e3b2d9`, forking the chain into two heads.
- **Fix:** re-point `source_ranges` after the coedit migration. Single linear chain `c7a1f0e3b2d9 -> b3e8f5a90c27 -> d8b2f1a05c93`. The `source_ranges -> provenance_ledger` FK still resolves, since `c7a1f0e3b2d9` runs first.
- **Safe:** nothing is deployed yet, so relabeling a merged migration cannot desync a live `alembic_version`.

Verified: `alembic heads` shows one head, upgrade + downgrade run clean, and the `init_db`-backed suites pass again.

## Additional Options
- [x] Override Linear Check

### DIFF
--- a/backend/app/db/migrations/versions/d8b2f1a05c93_source_ranges.py
+++ b/backend/app/db/migrations/versions/d8b2f1a05c93_source_ranges.py
@@ -8,7 +8,7 @@ inspector because ``0001_initial`` builds fresh databases from the current
 models.
 
 Revision ID: d8b2f1a05c93
-Revises: c7a1f0e3b2d9
+Revises: b3e8f5a90c27
 Create Date: 2026-07-17 00:00:00.000000+00:00
 """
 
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
Confirming the analysis from the duplicate #463 (same fix, byte-identical re-chain): #446 and #460 merged with migrations both revising c7a1f0e3b2d9, so main has two alembic heads — every fresh boot and every PR's backend-test CI fails on 'Multiple head revisions'. Re-pointing is safe: no environment has applied the two in conflicting order (dev-wiki hasn't applied either yet; CI schemas are ephemeral). Either PR unblocks; happy to close #463 in favor of this one.

Bo-Onyx:
#463 merged — this one can be closed as superseded. Thanks for jumping on it!

PR #465, opened 2026-07-20
feat(wiki): content-span read endpoint, map page spans to their source documents

nmgarza5:
## Description

The read side behind in-document source highlighting (the "this paragraph came from this Google Doc" part of source attribution).

- `GET /api/wiki/source-spans?path=` returns the live `source_ranges` for a page joined to their ingest source facts (`SourceSpan`: offsets + document id/type/url/title), ordered by offset.
- Spans are remapped to HEAD before reading, so offsets always line up with the page's current body (same backstop pattern as the comments GET path).
- Read-gated with the same ACL check as reading the page body.

Backend only. The FE highlighter that consumes this comes in the frontend phase.

## Additional Options
- [x] Override Linear Check

### DIFF
--- a/backend/app/api/wiki.py
+++ b/backend/app/api/wiki.py
@@ -80,7 +80,7 @@ from app.wiki import (
 )
 from app.ingest import settings as ingest_settings
 from app.models.update_policy import UpdateHealthResponse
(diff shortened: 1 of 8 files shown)


nmgarza5:
@greptile

PR #471, opened 2026-07-21
refactor(automanage): rename API to /api/automanage + schedule selector

Bo-Onyx:
Stacked on #470 (base = `bo/automanage-runner-killswitch-guard`; retarget to `main` after #470 merges).

Scope: **API rename + the frontend it requires.** The task-queue rename and the pod-consolidation work are a **separate follow-up PR** — deliberately not here.

## Backend — API-layer rename only
The router serves the whole Wiki Auto Management surface (settings, sweeps, proposals, execution), so `/api/detection` undersold it.
- `/api/detection` → `/api/automanage`; `app/api/detection.py`, `app/models/detection.py` → `automanage.py`.
- The **task queue stays `detection`** — its rename (and the pod-consolidation we discussed) is a separate PR. Queue/task/worker/compose/helm are untouched here.
- Backend + frontend ship together in this PR, so **no compatibility alias** for the old prefix is needed.
- API test URLs updated; queue/task refs untouched.

## Frontend
- `autoOrganize.ts` / `swr-keys` repointed `/detection/*` → `/automanage/*`; added the `schedule` field. `updateAutoOrganizeSettings(patch)` replaces the enabled-only helper.
- `/admin/auto-organize` gains an **Off / Daily / Weekly** schedule selector next to the kill switch, disabled while the switch is off. Seeds the SWR cache from the PUT response (`revalidate:false`), same pattern as the toggle.

## Test plan
- ruff + basedpyright clean.
- Automanage/detection suite green (68); API tests green without the alias (18); frontend typechecks clean (two unrelated `SvgHome`/`SvgSquarePlus` icon errors are pre-existing on `main`).
- No queue/stream rename here, so the orphaned-backlog deploy concern does not apply.
Screenshot 2026-07-20 at 10 33 42 PM


Bo-Onyx:
@greptile

PR #472, opened 2026-07-21
feat(automanage): schedule selector + repoint frontend to /api/automanage

Bo-Onyx:
Stacked on #471 (base = `bo/automanage-rename-queue-split`; retarget to `main` after #471 merges). Completes **task #3**'s UI.

Two things, both required once the backend API is renamed:

1. **Repoint the frontend** off the old `/api/detection` namespace → `/api/automanage` (`autoOrganize.ts`, `SWR_KEYS.autoOrganizeSettings`). Without this the admin page breaks after #471, so merge this right after.
2. **Schedule selector** — an Off / Daily / Weekly segmented control (Opal `Button`s) next to the existing kill switch on `/admin/auto-organize`, disabled while the master switch is off. Daily/weekly run at a fixed off-peak time (08:00 UTC / Mondays) per the backend.

Also: `updateAutoOrganizeSettings(patch)` replaces the enabled-only helper — a partial PUT that carries both `enabled` and `schedule`. Both the switch and the selector seed the SWR cache from the PUT response (`revalidate: false`), so a rapid second change can't be clobbered by a stale in-flight GET (same fix as the earlier toggle review).

## Test plan
- `bun run typecheck` clean for the changed files. (Two unrelated `SvgHome`/`SvgSquarePlus` icon errors are pre-existing on `main` from a local Opal drift — not touched here.)
- Toggle off → schedule selector + sweep disabled. Toggle on → pick Daily/Weekly/Off; selection persists across reload.


Bo-Onyx:
Superseded by #471 — the frontend repoint + Off/Daily/Weekly schedule selector were folded into #471 (which merged). Closing; the queue-pod consolidation we discussed will get its own fresh PR.

PR #474, opened 2026-07-21
refactor(tasks): group worker queues into 2 pods

Bo-Onyx:
Stacked on #473 (base = `bo/automanage-queue-groups`; retarget to `main` after #473's rename merges). This is the **grouping** half we split out of #473.

**Queues are the isolation unit; worker processes are the memory unit.** A queue is cheap (a Redis stream + its own consumer/thread pool); a worker pod pays a ~185 MiB import-graph floor regardless of how many queues it serves. So one process can host several queues — each its own consumer — keeping backlogs isolated (separate streams; I/O-bound work releases the GIL) while paying the floor once.

- `run_worker` now takes **one or more** queue names, spawns one `run_consumer` (thread pool) per queue in its own thread, binds one metrics server (lowest port among its queues), and installs signal handlers once on the main thread (new public `queue.install_signal_handlers`) so one SIGTERM drains every consumer.
- **Deployment: 5 pods → 2.** `coedit` — the only real-time, user-visible-freshness path — keeps its own pod; `documents`/`triggers`/`automanage`/`lightweight_maintenance` share a `background` pod. docker-compose has two worker services; helm switches per-queue → `worker.groups` (per-group Deployment, `metricsPort` per group).
- Per-queue concurrency and isolation unchanged.

Verified locally (before the rename/grouping re-slice): 2 worker pods up, `worker-background` runs all four consumers at correct concurrency, an admin sweep runs end-to-end on the grouped `automanage` consumer; **~558 MiB across 2 pods vs ~985 MiB across the old 5 (~430 MiB saved)**; `helm template` renders exactly 2 worker Deployments.

Doctrine + the (still-planned) `automanage_nearline`/`automanage_offline` split are documented on the wiki: `Engineering Projects/Agent Wiki Project/design/Queues and Workers.md`.

## Test plan
- ruff + basedpyright clean; automanage/queue suite green; `docker compose config` parses (2 worker services); `helm template` renders 2 worker Deployments.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -251,11 +251,12 @@ working dir, identity, and commit-on-write. Path validation goes through
 ### Background work — Redis Streams queues, not threads
 
 If something might take more than ~100ms, queue it. Tasks live under
-`app/tasks/` and bind to one of four `TaskQueue` instances in
+`app/tasks/` and bind to one of five `TaskQueue` instances in
 `app/tasks/queues.py` — `documents_queue` (LLM doc-reconciliation),
 `triggers_queue` (NL trigger eval, delta + scheduled), `coedit_queue`
 (co-edit session checkpoints — commits a live buffer, AI-merges on a
-concurrent commit), or `lightweight_maintenance_queue` (sub-second
+concurrent commit), `automanage_queue` (Wiki Auto Management — whole-space
+sweeps + proposal execution), or `lightweight_maintenance_queue` (sub-second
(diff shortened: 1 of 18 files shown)


Bo-Onyx:
Folded into #473 — the grouping now lands together with the rename+split so the queue split never bumps the worker count (5→2, not 5→6→2), and the heavy/light blast-risk split supersedes the coedit/background grouping this PR had. Closing.

PR #489, opened 2026-07-23
feat(coedit): migrate live-session transport from SSE+HTTP to WebSocket

raunakab:
## Summary
- Pure transport swap for the co-edit live session: the 8 REST/SSE endpoints (`join`/`leave`/`op`/`cursor`/`checkpoint`/`session`/`ops`/`stream`) collapse into one `WebSocket` route that multiplexes the same JSON message shapes over the same domain logic (`app/wiki/coedit.py` untouched). CodeMirror, the checkpoint/git pipeline, and the Postgres LISTEN/NOTIFY broadcast fan-out are all unchanged.
- Write permission is re-checked on every inbound message, not just at connect, matching the old per-POST `require_can` exactly.
- No client-sent `leave` message — the server's disconnect handler is the sole leave signal, firing on any connection loss (more reliable than the old keepalive-fetch-on-unload approach).
- Two follow-up bugs found via manual testing and fixed:
  - `FileView.tsx`: the template picker briefly flashed on real (non-blank) pages, and the loading spinner rendered unstyled top-left instead of centered.
  - A frontend main-thread freeze: `components.tsx`'s pre-existing op-retry loop had no backoff, which was safe under the old `fetch()`-based client (real network I/O rate-limited it) but spun as an unbroken microtask chain once `svc.ts`'s WS-based `request()` could reject synchronously on a closed socket.
  - A backend thread-pool exhaustion bug in `_send_loop`'s heartbeat poll (orphaned threads on cancellation, tightened poll interval to bound the damage).

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -135,21 +135,21 @@ def _handle_op(outbox: queue.Queue[coedit_channel.QueueItem], session_id: int, u
             client_id=msg.client_id,
         )
     except _WsActionError as e:
-        outbox.put_nowait(OpResultFrame(request_id=msg.request_id, ok=False, error=e.error).model_dump())
+        outbox.put_nowait(OpResultFrame(request_id=msg.request_id, ok=False, error=e.error).model_dump(by_alias=True))
         return
     except ValueError:
         outbox.put_nowait(
-            OpResultFrame(request_id=msg.request_id, ok=False, error="invalid_op").model_dump()
+            OpResultFrame(request_id=msg.request_id, ok=False, error="invalid_op").model_dump(by_alias=True)
         )
         return
     if out is None:
         outbox.put_nowait(
-            OpResultFrame(request_id=msg.request_id, ok=False, error="stale_version").model_dump()
+            OpResultFrame(request_id=msg.request_id, ok=False, error="stale_version").model_dump(by_alias=True)
         )
         return
     coedit.touch(session_id, user.id, edited=True)
     outbox.put_nowait(
-        OpResultFrame(request_id=msg.request_id, ok=True, version=out.version).model_dump()
+        OpResultFrame(request_id=msg.request_id, ok=True, version=out.version).model_dump(by_alias=True)
     )
     # caret_seq rides the frame so peers render the author's caret at the
     # edit; None (cleared caret / legacy client) = no caret assertion.
@@ -191,10 +191,10 @@ def _handle_checkpoint(outbox: queue.Queue[coedit_channel.QueueItem], session_id
     try:
         _require_active(session_id, user, "write")
     except _WsActionError:
-        outbox.put_nowait(CheckpointResultFrame(request_id=msg.request_id, ok=False).model_dump())
+        outbox.put_nowait(CheckpointResultFrame(request_id=msg.request_id, ok=False).model_dump(by_alias=True))
         return
     checkpoint_coedit_session(session_id)
-    outbox.put_nowait(CheckpointResultFrame(request_id=msg.request_id, ok=True).model_dump())
+    outbox.put_nowait(CheckpointResultFrame(request_id=msg.request_id, ok=True).model_dump(by_alias=True))
 
 
 def _handle_get_ops(outbox: queue.Queue[coedit_channel.QueueItem], session_id: int, user: User, raw: dict[str, Any]) -> None:
@@ -205,7 +205,7 @@ def _handle_get_ops(outbox: queue.Queue[coedit_channel.QueueItem], session_id: i
         outbox.put_nowait(
             OpsResultFrame(
                 request_id=msg.request_id, ok=False, error=e.error, current_head_version=0, ops=[]
-            ).model_dump()
+            ).model_dump(by_alias=True)
         )
         return
     # Head version + ops read as one consistent snapshot (see
@@ -227,7 +227,7 @@ def _handle_get_ops(outbox: queue.Queue[coedit_channel.QueueItem], session_id: i
                 )
                 for r in result.ops
             ],
-        ).model_dump()
+        ).model_dump(by_alias=True)
     )
 
 
@@ -365,7 +365,7 @@ async def ws(websocket: WebSocket, path: str, user: User = Depends(require_user_
                 version=sess.version,
                 base_sha=sess.base_sha,
                 participants=participants,
-            ).model_dump()
+            ).model_dump(by_alias=True)
         )
 
         recv_task = asyncio.create_task(_recv_loop(websocket, conn.queue, sess.id, user))
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -253,17 +253,39 @@ async def _recv_loop(websocket: WebSocket, outbox: queue.Queue[dict[str, Any]], 
             log.warning("coedit ws: malformed %r message", msg_type)
 
 
+# How often _send_loop's blocking drain call returns to check for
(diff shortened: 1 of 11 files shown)


backend/app/api/coedit.py
Bo-Onyx, line 1: First asyncio in the backend — unavoidable for WS, but let's scope the exception: note in the docstring that asyncio is transport-only, everything below the route stays sync via to_thread. Please add a matching CLAUDE.md rule.

backend/app/api/coedit.py
```diff
@@ -98,202 +86,281 @@ def _checkpoint_if_last_left(session_id: int) -> None:
...
+# thread pool every asyncio.to_thread call in the process draws from. Under
+# frequent reconnects, orphaned threads accumulate faster than they drain,
+# eventually exhausting the pool — at which point *every* asyncio.to_thread
+# call across the whole backend (new connects, every op/cursor/checkpoint
+# message, for every session) queues for a free worker. That reads as "the
+# whole app froze," and can plausibly cause more disconnects itself (a
+# starved event loop can miss uvicorn's own ping/pong window and get
+# closed as unresponsive) — a self-reinforcing spiral. Bounding the poll to
+# ~1s bounds any orphaned thread's worst case to ~1s instead of 15.
+_SEND_LOOP_POLL_SECONDS = 1.0
```
Bo-Onyx, line 275: The 1s poll bounds the damage, but orphans still eat the process-wide default executor shared by every to_thread call. Fine to ship; follow-up: a dedicated bounded executor for this route, or an interruptible drain (sentinel on cancel) so orphans never form. I'll file an issue.

backend/tests/test_coedit_api.py
```diff
@@ -141,195 +176,286 @@ def test_op_stamps_last_edited_at(client):
...
-        "/api/coedit/op",
-        json={"session_id": sid, "base_version": 0, "changes": [{"from": 0, "to": 0, "insert": "x"}]},
+    with _ws(client) as (_owner_ws, _joined):
+        login_fastapi(client, other)
+        with _ws(client) as (ws, _joined2):
+            result = _op(ws, 0, [{"from": 0, "to": 0, "insert": "x"}])
+            assert result == {"type": "op_result", "request_id": "r", "ok": False, "version": None, "error": "forbidden"}
+
+
+def test_write_permission_revoked_mid_session_is_enforced_on_next_message(client):
```
Bo-Onyx, line 404: This test pins a behavior we haven't actually committed to: _require_active re-runs require_can on every op/cursor/checkpoint message, so revoking write mid-session takes effect on the next message without a reconnect. The planned Yjs transport gates differently — pycrdt-websocket rooms apply updates as they arrive, and whether there's a per-update hook to re-check permissions is still unvalidated. So before that endpoint exists, we should decide which this is: (a) a hard invariant the Yjs migration must preserve (then this test's behavior becomes an acceptance criterion for the spike), or (b) connect-time gating is acceptable and this PR is just incidentally stronger. Either is fine — I just don't want (a)-vs-(b) discovered as a silent regression during cutover.

Bo-Onyx (commented):
LGTM — clean transport swap, domain logic untouched, and the WS infra carries over to Yjs. Three non-blocking notes inline.

backend/app/api/coedit.py
```diff
@@ -98,202 +86,281 @@ def _checkpoint_if_last_left(session_id: int) -> None:
...
+        for t in done:
+            exc = t.exception()
+            if exc is not None and not isinstance(exc, WebSocketDisconnect):
+                raise exc
+    except WebSocketDisconnect:
+        pass
+    finally:
+        coedit_channel.disconnect(conn.id)
+        await asyncio.to_thread(_disconnect_sync, sess.id, user.id)
+        log.info("coedit ws closed session=%s user=%s", sess.id, user.id)
```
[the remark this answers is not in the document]
raunakab: Fixed in 1ec417a — moved `accept()`/`conn` assignment inside the `try`, with a `conn is not None` guard in `finally`, matching your proposed fix exactly.

frontend/src/lib/editor/svc.ts
```diff
@@ -1,137 +1,282 @@
...
+      }
+      pending.clear();
+      if (!joined) {
+        // The join handshake itself failed — mirrors the old client's
+        // "no read-only fallback" contract: surface it, don't retry silently.
+        reject(
+          event.code === 1008
+            ? new ApiError(403, "You don't have permission to edit this page.")
+            : new ApiError(0, "Failed to join the editing session."),
+        );
```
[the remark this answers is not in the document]
raunakab: Confirmed and fixed in 1ec417a — dropped the code === 1008 branch entirely. You're right that it's unreachable: browsers report the generic 1006 for any pre-upgrade handshake failure, and there's no reliable signal here to distinguish "forbidden" from any other pre-accept failure. Now always shows the generic "Failed to join the editing session." message.

frontend/src/lib/editor/svc.ts
```diff
@@ -1,137 +1,282 @@
...
+ * which depended on the client successfully transmitting during teardown.
+ * `closeSession` is the equivalent — it just closes the connection.
  *
  * Offsets are UTF-16 code units — which is exactly what JS string indexing and
  * `slice` use, so the ops interoperate with the server
  * (`Change` in `app/wiki/coedit.py`) without conversion; see `utils.ts` for
  * `diffToChange`/`applyChange`.
  */
-import { apiFetch, apiStream } from "@/lib/api";
+import { apiSocketUrl, ApiError } from "@/lib/api";
```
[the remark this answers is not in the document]
raunakab: Fixed in 1ec417a — went through svc.ts, hooks.ts, coedit_channel.py, and api/coedit.py's module docstring and rewrote every comment/docstring you flagged to describe current behavior instead of contrasting against the old SSE+HTTP design.

backend/app/api/coedit.py
raunakab: Done in 25ee286 — added the scoping note to coedit.py's module docstring (async is transport-only, everything below the route is plain sync via to_thread) and a matching "WebSocket routes" rule in CLAUDE.md, including the concurrent-writer gotcha we hit in the same file.

backend/app/api/coedit.py
raunakab: Went ahead and fixed the root cause directly in 4dfaa73, rather than shipping the 1s mitigation and filing a follow-up — added coedit_channel.wake(conn_id), which pushes a CLOSE_SIGNAL sentinel into the connection's own queue right before cancelling its send loop. That unblocks a blocked drain() call immediately (same mechanism as a real frame arriving), instead of leaving it to run out its poll timeout. No orphan window left to bound — no follow-up issue needed for this specific gap. The dedicated-executor idea is still worth keeping in your back pocket as containment for any *other* to_thread caller that hits this same class of bug, but this route no longer produces orphans at all.

backend/tests/test_coedit_api.py
raunakab: Decided: (a) — hard invariant. Connect-time-only gating (what the Yjs/pycrdt-websocket work currently does) is a real, if narrow, security regression: a mid-session write-permission revocation wouldn't take effect until the client reconnects. Keeping test_write_permission_revoked_mid_session_is_enforced_on_next_message as-is — its behavior (re-check require_can on every write message, not just at connect) is now an acceptance criterion the Yjs migration needs to satisfy before it merges, not just this PR's incidental behavior. Flagging this explicitly so it doesn't get silently discovered as a gap during cutover, per your original ask.

raunakab:
@greptile-apps, please re-review.

PR #490, opened 2026-07-23
feat(wiki): hubspot and braintrust icons in the sources map

nmgarza5:
## Description

Follow-up to #487: the dev wiki's live ingest traffic includes `hubspot` and `braintrust` documents, which fell back to the generic file icon. Adds the existing Opal logos to the sources icon map.

## Additional Options
- [x] Override Linear Check


nmgarza5:
@greptile

PR #502, opened 2026-07-23
feat(automanage): folder-chain flattening detector (Wave 2)

Bo-Onyx:
First Wave-2 detector: single-child folder-chain flattening.

## What it detects
A **chain** is a run of folders each containing exactly one subfolder (a `.gitkeep` marker aside); the **tail** is the first folder with real content. Every level between head and tail is a pass-through. The proposal flattens the chain by moving the tail's pages up into the head, preserving structure below the tail (`head/link/tail/sub/x.md` → `head/sub/x.md`), enumerating every concrete move pair — the executor's exact-path scope rail requires concrete paths.

## Division of labor
- The agentic applier **only moves pages** (the move-must-preserve-content rail from #500 holds it to that). Emptied wrappers keep their `.gitkeep` markers and are deliberately left behind — the **empty-folder detector** proposes their deletion on a later pass. Detectors compose through wiki state, not through each other.
- An entirely empty chain (no pages) stays the empty-folder detector's case.

## Proposal shape
Chain folders ride in `source_paths` (head first, tail last) ahead of the page paths: they're the subject of the op, they widen the applier's path scope to the wrappers, and `validate()` re-derives the chain from them robustly instead of guessing it back out of page paths. `target_paths` = destination pages, index-aligned.

## Premise validation (execute-time)
Stales when: the chain is no longer single-child down to the same tail; the page set under the tail changed (a page added since approval would be stranded half-flattened); or a destination is occupied — judged **case-insensitively**, same as at detect time, so a flatten can never manufacture a case collision. Body edits keep the proposal valid (intent-level consent).

## Guards
- 7-day quiet window on the head — a structure someone is actively building (this month's folder just went in) is left alone.
- Non-page content under the tail (folder-scoped triggers, attachments) skips the chain.
- A chain nested inside an accepted chain's tail is skipped this run — the outer flatten recreates it under the head and a later pass re-detects it.
- **Never auto-approvable**: which folder name survives is naming judgment; a human confirms even in AI-managed scopes.

`move` joins `SUPPORTED_OPS` (the one-line policy decision the allowlist exists for) — the agentic applier already executes moves; all rails (scope, targets-survive, move-preserves-content) apply unchanged.

Note: the Path-2 review banner's detail line says "Keeps X — retires Y" for every target-bearing op, which misreads for `move` — small stacked frontend PR to follow.

13 new tests; automanage/detector/executor suite green (109 passed); ruff + basedpyright clean.

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -6,7 +6,7 @@ the originating conversation for the design rationale.
 
 The first ``asyncio`` in this backend, scoped deliberately: ``async`` here
 covers connection lifecycle only (accept, the recv/send loops, task
-orchestration) — every ``_connect_sync``/``_disconnect_sync``/``_handle_*``
+orchestration) — every ``_connect_sync``/``record_leave``/``_handle_*``
 helper below is a plain sync function, run via ``asyncio.to_thread`` from
 the loops, with no idea it's being called from an async context at all. See
 CLAUDE.md's "WebSocket routes" rule before adding another route like this
@@ -31,7 +31,6 @@ from __future__ import annotations
 import asyncio
 import logging
 import queue
-import threading
 from typing import Any
 
 from fastapi import APIRouter, Depends, WebSocket, WebSocketDisconnect
@@ -52,6 +51,7 @@ from app.models.coedit import (
     ParticipantOut,
 )
 from app.tasks.coedit_checkpoint import checkpoint_coedit_session
+from app.tasks.coedit_leave import leave_coedit_session, record_leave
 from app.wiki import coedit, coedit_channel, git
 
 router = APIRouter()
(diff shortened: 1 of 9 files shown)


Bo-Onyx:
Addressed in the latest push. The executor now recovers the moves the run actually performed from the pre-minted doc ids (a source whose id row is live at a different path was moved there), captured before the revert, and replays them **in reverse through `notify.after_path_move`** — the same single chokepoint the forward move used. So ACL/owner rows, update policies, doc ids, co-edit sessions, comments, the trigger cache, and the search index all follow the page back, with no per-table unwinding code. A freshly written destination that the revert deletes (a write, not a move) gets its search entry purged as a backstop; ACL/policy orphans at such paths remain the known-debris class the move machinery heals on collision.

Regression test: a scripted run lands one move then fails — after the revert, the doc id, owner row, and search index all point at the source again and nothing references the removed destination.

PR #504, opened 2026-07-23
fix(automanage): make sweep-slot acquisition atomic

Bo-Onyx:
Follow-up to #501, resolving the review comment: the guard was check-then-insert (`running_sweep_exists()` in one transaction, the run-row insert in another), so the singleton held only because the queue has a single consumer — direct callers or a multi-consumer deployment could still double-scan.

- **The running sweep row is now the slot.** A partial unique index — `detection_runs(trigger) WHERE status = 'running' AND trigger = 'sweep'` — allows at most one, so acquisition is a single guarded `INSERT … ON CONFLICT DO NOTHING RETURNING id`. The id comes back only when the insert won; no exists-check window at any concurrency. (`RETURNING` rather than rowcount — rowcount is unreliable for `ON CONFLICT` inserts under implicit returning; a test caught it returning truthy for a losing insert.)
- **Corpses are failed over, not ignored.** A `running` row older than `STUCK_RUN_MAX_AGE_HOURS` is marked `failed` (with an explanatory error) in the same transaction before the insert, so it can't hold the slot forever and the run history shows what happened.
- **Acquisition sits inside `run_detection`,** after the kill-switch check — a disabled run never leaks a `running` row, and every sweep caller (manual API trigger, crons, direct calls) inherits the guard rather than only the `run_sweep` wrapper.

Migration `d5e8a1c4f7b2` adds the index (`if_not_exists`, since `0001_initial` materializes model metadata on fresh installs).

Tests: overlap skipped; first-wins acquisition asserted directly at the repo layer; corpse failover verified to close the row; sequential sweeps unaffected. Suite green; ruff + basedpyright clean.

### DIFF
--- a/backend/app/db/migrations/versions/d5e8a1c4f7b2_sweep_singleton_index.py
+++ b/backend/app/db/migrations/versions/d5e8a1c4f7b2_sweep_singleton_index.py
@@ -22,6 +22,30 @@ depends_on: str | None = None
 
 
 def upgrade() -> None:
+    # Databases from the check-then-insert era can hold several ``running``
+    # sweep rows (crashed workers never marked failure) — the unique index
+    # can't be created over them. Reconcile first: keep the newest, fail the
+    # rest, mirroring the runtime corpse failover. If the survivor is a live
+    # sweep, its own completion overwrites the status later; if it's a
+    # corpse, try_start_sweep's age cutoff clears it on the next sweep.
+    op.execute(
+        sa.text(
+            """
+            UPDATE detection_runs
+            SET status = 'failed',
+                error = 'superseded — duplicate running sweep reconciled by migration',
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
Addressed in the latest push. The migration now reconciles before creating the index: every `running` sweep row except the newest is flipped to `failed` with an explanatory error — the same semantics as the runtime corpse failover, applied once. If the surviving row is a genuinely live sweep, its own completion overwrites the status later; if it's a corpse, `try_start_sweep`'s age cutoff clears it on the next sweep.

Verified empirically against a simulated legacy database (index dropped, three `running` sweeps + one `completed` seeded): upgrade succeeds, the two older rows are failed over, the newest stays `running`, the completed row is untouched, and the index lands. Fresh databases can never hit this — `0001_initial` materializes the index from model metadata, so duplicates are impossible from birth there.

Bo-Onyx:
@greptile

PR #509, opened 2026-07-24
fix(coedit): don't lose the leave when the WS task is cancelled

Bo-Onyx:
Fixes the CI failures hitting unrelated PRs (#502, #504) since the WS transport migration (#489).

## The bug
The disconnect handler is **the** leave signal — there is no client leave message. It ran as `await asyncio.to_thread(_disconnect_sync, …)` in the endpoint's `finally`. When the endpoint task is *cancelled* (server shutdown; the test portal tearing down the connection), every `await` inside the cancelled task re-raises `CancelledError` instead of completing — so the leave silently never ran: a phantom participant forever, and a dirty buffer that never got its last-leave checkpoint. Production impact: a server shutdown mid-session drops the checkpoint the same way.

## Evidence
`test_disconnect_of_last_participant_checkpoints` failing across branches (including a migration-only PR) on part of the runner fleet; an instrumented run showed `_disconnect_sync` never executed — session still open, participant still present after a 15s wait, no handler log lines — plus an orphaned `_recv_loop` task warning implicating the cancellation path.

## The fix
The `finally` now checks `task.cancelling()` up front: mid-cancellation, the leave is handed to a plain `threading.Thread` that no cancellation can touch; the normal disconnect path keeps awaiting so the leave is recorded before the handler returns. Checked **before** awaiting rather than caught after, to avoid double-running the leave — a cancelled `await asyncio.to_thread(...)` has already submitted its executor item by the time it re-raises.

## Tests
Both disconnect tests now wait for the handler's *effects* (participant gone; checkpoint committed + session closed) instead of assuming the handler finished by the time `websocket_connect.__exit__` returned — an ordering the API never promised. The checkpoint wait sits inside `immediate_mode` on purpose: once the flag drops, a late enqueue would land on the real queue and never run in the test.

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -31,7 +31,6 @@ from __future__ import annotations
 import asyncio
 import logging
 import queue
-import threading
 from typing import Any
 
 from fastapi import APIRouter, Depends, WebSocket, WebSocketDisconnect
@@ -327,10 +326,13 @@ def _connect_sync(path: str, user: User) -> coedit.SessionRow:
 def _disconnect_sync(session_id: int, user_id: str) -> None:
     # Only mark the user gone when their last connection closes, so a
     # second tab doesn't evict them.
-    if not coedit_channel.user_still_connected(session_id, user_id):
+    still = coedit_channel.user_still_connected(session_id, user_id)
+    log.info("coedit leave session=%s user=%s still_connected=%s", session_id, user_id, still)
+    if not still:
         coedit.leave(session_id, user_id)
         coedit_channel.broadcast_presence(session_id)
         _checkpoint_if_last_left(session_id)
+        log.info("coedit leave done session=%s", session_id)
 
 
 @router.websocket("/ws")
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
The unmanaged thread is gone as of the latest push — that branch turned out to be built on an unconfirmed theory anyway: the same code passed one CI run and failed the next on the *same* runner image, so this is a per-run race, not the clean cancellation story the thread was written for. The branch is currently a plain awaited `asyncio.to_thread` (the pre-existing managed shape) plus INFO tracing at each lifecycle stage (wait-returned, finally-entered + `cancelling()` value, leave still-connected check, leave-done, ws-closed), captured by the disconnect tests. The next failing run will show exactly how far the endpoint gets; the final fix will route any deferred cleanup through the task-queue boundary rather than a raw thread.

Bo-Onyx:
Root cause found — a traced failing run showed the endpoint's `finally` entered with `cancelling=1` and the leave never executing. The mechanism: a cancelled `await asyncio.to_thread(...)` has submitted its executor item, but cancellation propagates to the wrapped future, and **cancelling a not-yet-started `concurrent.futures.Future` discards it from the executor queue**. Fast machine → a pool worker grabs the item before the cancel lands → leave runs. Busy CI runner → cancel wins → leave silently dropped. Hence per-run nondeterminism on identical code, and why no up-front `cancelling()` check closes the window (the cancel can land mid-await).

Final shape, which also resolves this thread's concern about the unmanaged thread:
- Leave logic moved to `app/tasks/coedit_leave.py` (`record_leave` + a registered `coedit_queue` task) — the repository's managed background-work boundary, with the queue's logging/exception handling.
- Normal disconnect path unchanged: the handler awaits `record_leave` and the leave is recorded before it returns.
- Cancellation path: `except CancelledError` → **enqueue** `leave_coedit_session` — a synchronous Redis send nothing can cancel, durable across the shutdown that caused it — then re-raise. No raw thread, no shutdown delay, exceptions route through the task layer.
- `record_leave` is idempotent (delete-if-exists; checkpoint no-ops on clean), so the rare both-ran overlap is safe. On a worker the empty channel registry makes the still-connected guard resolve correctly.
- Probe logging removed; tests wait for effects under `immediate_mode`.

PR #510, opened 2026-07-24
feat(automanage): stub/placeholder-page detector

Bo-Onyx:
The last mechanical detection technique — closes out the deterministic wave entirely.

## What it detects
A page that never grew past its title. **Content bytes** = the body with heading lines, horizontal rules, and blank lines stripped, whitespace collapsed; at or under a deliberately low floor (24 bytes — catches "TODO" / "Coming soon.", stays well under one meaningful sentence) → `delete_page` proposal, once the page has been quiet for 7 days (any edit restarts the clock).

## Boundaries, on purpose
- **Byte-identical to a current template body → skipped**: template-echo owns that case, and its proposal names the template — a better review story than "this page is small".
- **Partially-filled template skeletons are NOT stubs**: telling "barely started" from "abandoned scaffold" takes judgment — that's the future fuzzy detectors' job, not a byte count's. The floor exists for pages that are *obviously* placeholders.
- **Never auto-approvable**: smallness is evidence, not proof; a terse page can be exactly as long as it needs to be.

## Execution
Zero executor changes — `delete_page` and the `trash_page` applier tool have existed since template-echo. `validate()` re-checks the premise at execute time: real content written since approval stales the proposal; scaffolding-only shuffles keep it valid.

10 new tests (content measurement, floor calibration against a real short sentence, template-echo precedence, quiet window, trigger gating, all three validate outcomes); automanage suite green (121 passed); ruff + basedpyright clean.

### DIFF
--- a/backend/app/wiki/automanage/detectors/__init__.py
+++ b/backend/app/wiki/automanage/detectors/__init__.py
@@ -12,6 +12,7 @@ from app.wiki.automanage.detectors import (
     case_collision,
     empty_folder,
     folder_chain,
+    stub_page,
     template_echo,
 )
 from app.wiki.automanage.detectors.base import (
@@ -27,6 +28,7 @@ DETECTORS: list[Detector] = [
     template_echo.DETECTOR,
     case_collision.DETECTOR,
     folder_chain.DETECTOR,
+    stub_page.DETECTOR,
 ]
 
 # Validation dispatch: a proposal records which detector authored it, and the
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
Fixed in the latest push: `validate()` now re-checks the quiet window in addition to the content floor — an edit after approval that stays under the floor restarts the 7-day clock and stales the proposal (activity is activity, whatever its size; the detector simply re-proposes once the page settles again). Test updated to pin both sides: a post-approval scaffolding edit stales via the quiet window, and the same state with the window elapsed keeps the premise valid.

PR #511, opened 2026-07-24
feat(admin): sweep outcome + run history on the Auto Organize page

Bo-Onyx:
"Run a sweep" previously said **"Sweep started."** and nothing else — the run-history endpoint (`GET /api/automanage/runs`, from #426) existed but nothing in the UI called it, so the admin who pressed the button never learned whether the sweep found 0 proposals or 15 (or failed) without curling the API or stumbling onto a page banner.

Frontend-only:

- **Sweep outcome**: the button watches the run history (2s SWR poll only while a triggered run is queued/running; the new run is recognized **by id against a pre-trigger snapshot** — the request only enqueues, so timestamp comparison against client time would be racy) and reports "Sweep finished — N proposals from M paths" or the failure error inline.
- **Recent sweeps card** under the controls: last runs with local time, manual/scheduled (from `triggered_by_user_id`), proposal + path counts, and status (failed runs show in the error color).

`lib/autoOrganize.ts` gains the `DetectionRun` type + `useDetectionRuns(poll)` hook via `SWR_KEYS.automanageRuns` per the swr-keys convention. Opal components throughout; typecheck clean.

First slice of the proposal-visibility work — the owner-notification design (dedup + impression discounting) remains the follow-up.
Screenshot 2026-07-24 at 10 52 49 AM

### DIFF
--- a/frontend/src/components/admin/AutoOrganize.tsx
+++ b/frontend/src/components/admin/AutoOrganize.tsx
@@ -176,31 +176,31 @@ const runColumns = (() => {
   return [
     tc.column("started_at", {
       header: "Started",
-      weight: 30,
+      weight: 32,
       enableSorting: false,
       cell: startedCell,
     }),
     tc.column("triggered_by_user_id", {
       header: "Trigger",
-      weight: 15,
+      weight: 16,
       enableSorting: false,
       cell: triggerCell,
     }),
     tc.column("proposals_emitted", {
       header: "Proposals",
-      weight: 20,
+      weight: 22,
       enableSorting: false,
       cell: proposalsCell,
     }),
     tc.column("paths_scanned", {
       header: "Scanned",
-      weight: 15,
+      weight: 18,
       enableSorting: false,
       cell: scannedCell,
     }),
     tc.column("status", {
       header: "Status",
-      weight: 20,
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
Fixed in the latest push. The trigger filter moved from the history table into `useDetectionRuns` itself, which now exposes only sweep rows — so both consumers (the outcome watcher and the "Recent sweeps" table) share one filtered source, and a future `on_create`/`on_write` run completing mid-watch can't masquerade as the sweep outcome or end the watch early. Server-side `?trigger=` filtering remains noted for the focused-trigger work, where a client filter over a limited fetch would start skewing page sizes.

Bo-Onyx:
Fixed in the latest push: the watch anchor is now snapshotted from an **awaited authoritative fetch** (`await refresh()`) before the sweep is enqueued, instead of from the possibly-unloaded SWR cache. After that await, an empty snapshot can only mean the history is *verifiably* empty — where any appearing sweep row genuinely is the new one — so a pre-existing terminal run can no longer masquerade as the requested sweep's outcome. The is-a-sweep predicate is shared (`sweepRuns`) between the hook and the snapshot path so the two can't drift.

PR #512, opened 2026-07-24
feat(editor): raw ProseMirror + Yjs live editor

raunakab:
## Summary

Replaces the wiki's live editor (Tiptap + OT-based buffer/version CAS) with **AgentWikiEditor**, built directly on `prosemirror-model`/`prosemirror-state`/`prosemirror-view` with `y-prosemirror` wired in by hand — no Tiptap dependency (open-core/paid-tier model was the reason to drop it). Live sync moves from the OT buffer store to a Yjs CRDT doc.

- **Backend**: `markdown_blocks`/`markdown_yjs`/`markdown_splice` — the markdown ⇄ Yjs codec, with a stable block-id scheme so untouched regions round-trip byte-for-byte on checkpoint. `coedit_ws.py` — a per-connection Yjs WS relay (`app/api/coedit.py`) with per-*update* (not just per-connect) write-permission re-checks, matching the invariant from #489. Checkpointing reuses the existing 3-way + AI merge gateway. Live-rebase (an external commit landing under an open session) now routes through a checkpoint+merge instead of attempting a CRDT-level fold. Comment/source anchors resolve across both commit-to-HEAD drift (existing `anchor_remap`) and HEAD-to-live drift (new `resolve_live_spans`).
- **Frontend**: hand-built schema/input-rules/keymap (`schema.ts`/`inputRules.ts`/`keymap.ts` — StarterKit's job, without StarterKit), y-prosemirror sync/cursor/undo plugins, and the comment/source highlight decoration plugin ported from the old CM6 field to a PM plugin.
- **Design doc**: `Engineering Projects/Agent Wiki Project/design/Co-Editing.md` in the wiki.

Also fixes a number of real bugs found during manual verification (browser testing against a live dev instance, not just typecheck): missing `markdown` CSS class (headings/lists/code/tables rendered unstyled), no base `prosemirror-view` stylesheet or focus-outline reset, no thematic-break input rule, `codeBlock`/`thematic_break`/`html_block`/`other` missing `group: "block"` (so they could never validly appear as a doc/blockquote/list child — and the latter three had no `toDOM`, so rendering one would throw), a Backspace-in-list bug (unconditional lift regardless of cursor position, then a re-wrap-into-new-list-item bug in the follow-up join), a migration that collided with `0001_initial`'s live `create_all` on a fresh database build, and a double-indented editor text bug (redundant padding stacked on top of centering).

## Test plan

- [x] `uv run ruff check app tests` / `uv run basedpyright` — clean
- [x] `bun run typecheck` / `bun run format:check` — clean
- [x] Manual verification in a real browser against a live dev backend: headings/lists/task-lists/code-blocks/tables/thematic-breaks create and render correctly; list Backspace lift/join behaves like Notion/Docs; per-update write-permission revocation closes the WS connection; two-tab live sync converges; checkpoint lands in git with untouched regions byte-stable
- [ ] Backend `pytest` suite — written (`test_markdown_blocks.py`/`test_markdown_yjs.py`/`test_markdown_splice.py`/`test_coedit_repo.py`/`test_coedit_checkpoint.py`/`test_coedit_rebase.py`/`test_coedit_ws_api.py`), not yet run — left for reviewer/CI per this repo's convention of not running tests mid-implementation


raunakab:
Closing — superseding this approach. The raw ProseMirror path turned out to require redefining far more than expected (round-tripping markdown alone took 10 review rounds on #513's codec layer), so we're switching to a Tiptap-based implementation instead, taking inspiration from the earlier `feat/onyx-editor` branch but starting fresh off current `main` rather than reusing either stale branch directly. Both `feat/agent-wiki-editor` and `feat/onyx-editor` are being renamed to `STALE/*` for reference.

PR #515, opened 2026-07-24
feat(automanage): proposal identity fields — dedup_key, doc_ids, revive history (schema)

Bo-Onyx:
Schema + field setup for the dedup model on the **Wiki Auto Management — Dedup** design page — split per review so the DB shape merges independently; the component that computes/consumes these lands in the stacked follow-up.

## Columns on `change_proposals` (migration `f4b7e2a9c1d5`, inspector-guarded)
- **`dedup_key`** (indexed) — the canonical identity string: `detector | op | doc-id terms | premise`. The cooldown scope is its content-free prefix — deliberately **not** a second column (derivable by construction; the follow-up queries it with `starts_with`).
- **`doc_ids`** — `{stable doc id: path at emit}`: which *documents* the proposal is about, keyed by the stable term so nothing can index it by a since-moved path; `base_shas`'s identity twin. Backs future per-document queries (JSONB key-existence).
- **`revive_count`** (default 0) + **`last_emitted_at`** — revival history; "keeps coming back" is not reconstructable after the fact.

## Draft-side field
`ProposalDraft.premise` — the detector-owned identity term (identity-side twin of `validate()`): body-dup stamps the shared blob, template-echo the matched template body; structural detectors none.

All inert in this PR: `create()` persists the fields; nothing computes or reads them yet. Snapshot semantics throughout — resolved at emit while path↔id bindings are true, never updated after (same philosophy as `base_shas` / `acl_fingerprint_before`).

Roundtrip tests only; behavior tests ride with the component. Follow-up: **#516** (Deduper: carry/reject-forever/revive/subject-cooldown + runner wiring).


Bo-Onyx:
Evolved during design review with Bo (three commits since the description):

- **`doc_ids` column** — the identity snapshot first-class, **keyed by doc id** with the emit-time path as label (a path-keyed map invites indexing by a since-moved path; id-keyed also makes "proposals about doc X" a JSONB key-existence query for the future overlap prefilter).
- **`emit_count` / `last_emitted_at`** — revival otherwise erases its own history; "keeps coming back" feeds the reviewer surface and detector-precision stats and is not reconstructable later.
- Docstring: key strings make detector names and op values **durable identifiers** — renames require a data migration over stored keys.

Schema-stability audit against the specced roadmap (reconciliation/selection, impressions+digest, variant gate, ranking): the remaining foreseeable changes to this table are one optional GIN index on `doc_ids` and at most two nullable columns whose semantics aren't designed yet; per-user notification state deliberately goes to its own table.

PR #516, opened 2026-07-24
feat(automanage): dedup component — carry, reject-forever, revive, cooldown

Bo-Onyx:
The dedup behavior on top of #515's identity fields — **pure identity**, per review: the post-rejection cooldown was pulled back out (it's a *selection*-step concern per the Dedup design page — "still detected and persisted, but unselectable" — and lands with the reconciliation/selection work, not as a dedup skip).

## `app/wiki/automanage/dedup.py`
A per-run `Deduper` decides each draft against the ledger by exact identity (`dedup_key` = detector | op | doc-id terms | premise):

- **CREATE** — new finding, persist (keys + `doc_ids` stamped).
- **SKIP_LIVE** — a pending/approved/applied row owns the finding: carried.
- **SKIP_REJECTED** — same ask was declined: never again.
- **REVIVE** — a stale/expired row returns to pending, whole operational face refreshed from the re-detected draft (**paths included** — identity followed the documents across moves, so the row must describe them where they live now), `revive_count`/`last_emitted_at` updated. Stale-from-approved revives to *pending*. Guarded transition; racing approvals lose cleanly.

## Wiring
Runner: `decide()` gates emit; REVIVE counts as emitted. Legacy path-based draft key renamed `legacy_dedupe_key` (one letter from the column otherwise) — computed, never stored — and kept as a transitional belt so pre-`dedup_key` rows keep blocking; retires by code deletion once the ledger drains or a one-shot key backfill lands.

## Verified
11 behavior tests (identity across renames via doc ids, reserved-name fallback, reject-forever, new-premise-after-rejection creates at dedup level, revival with same id + refreshed paths/anchors at unit and real-sweep level, applied-blocks-re-emit, cross-detector separation). Live-verified on the dev stack: emit → carry (0) → stale → revive (same id, refreshed run) → reject → silent forever.

Future slots: selection/invalidation + the subject cooldown reuse these keys (`starts_with` on the `dedup_key` prefix); the rejected-variant LLM gate is one more `DedupAction` branch.

### DIFF
--- a/backend/app/db/migrations/versions/b8d3f6a1c9e7_stale_unkeyed_pendings.py
+++ b/backend/app/db/migrations/versions/b8d3f6a1c9e7_stale_unkeyed_pendings.py
@@ -1,0 +1,67 @@
+"""Retire path-dedup-era pendings; make dedup_key unique.
+
+Revision ID: b8d3f6a1c9e7
+Revises: f4b7e2a9c1d5
+
+Rows created before dedup_key existed can't be matched by the identity
+dedup component; a pending one would get a keyed duplicate minted beside
+it on the first post-deploy sweep. Stale them instead — the sweep
+re-detects whatever is still true and re-creates it under identity keys
+(stale is the designed retry path). Approved/applied/rejected rows are
+left alone: approvals execute momentarily, and history is history.
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = "b8d3f6a1c9e7"
+down_revision: str | None = "f4b7e2a9c1d5"
+branch_labels: str | None = None
+depends_on: str | None = None
+
+
+def upgrade() -> None:
+    # One finding = one row, structurally (partial: legacy NULL keys exempt).
+    # Replaces the plain index from f4b7e2a9c1d5; if_not_exists because
+    # 0001_initial materializes model metadata on fresh installs.
+    op.drop_index(
+        "idx_change_proposals_dedup_key",
+        table_name="change_proposals",
+        if_exists=True,
+    )
+    op.create_index(
+        "ux_change_proposals_dedup_key",
+        "change_proposals",
+        ["dedup_key"],
+        unique=True,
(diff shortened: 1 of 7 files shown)


Bo-Onyx:
Valid catch — fixed in the latest push. `revive()` now refreshes the full operational face from the re-detected draft, **paths included**: identity followed the documents across the move (doc ids), so the revived row describes them where they live now — `source_paths`/`target_paths` and the `base_shas` keys stay aligned, the banner shows real paths, and the executor's per-path anchor lookups hit. New test pins the move-then-revive scenario end to end.

Bo-Onyx:
Two more review-driven changes folded into the commit:

1. **`legacy_dedupe_key` is gone entirely** — instead of carrying the transitional path-based belt, a data migration stales the (pre-key) pending rows with reason "superseded — re-detected under identity dedup keys"; the next sweep re-creates whatever is still true under proper keys (stale is the designed retry path, and no notification system exists yet, so re-keying costs nothing). Rejected/applied legacy history is left untouched. The old property, `taken_dedupe_keys`, and the runner belt are deleted — no "temporary" code to fossilize.

2. **One finding = one row is now structural** — `dedup_key` gets a partial unique index (NULL-exempt for legacy rows), same philosophy as the sweep-slot index: a duplicate insert (a bug, or a future concurrent-trigger race) fails loudly instead of silently forking a finding's history. Test pins the IntegrityError.

Bo-Onyx:
@greptile

Bo-Onyx:
Accurate observation, deliberate behavior — documented at the decision site now. Dedup is pure identity: a new premise on recently rejected pages CREATEs immediately **by design** (pinned by `test_new_premise_is_a_new_finding_even_after_rejection`). The post-rejection cooldown was deliberately moved out of dedup: per the Dedup design page it's a *selection*-step semantics ("detected and **persisted**, but unselectable") — a dedup-level skip couldn't even express it, since skipping doesn't persist. It ships with the reconciliation/selection PR. Interim exposure is bounded: triggering it requires the same pages to re-acquire a *different* identical-content premise right after a rejection, and each new premise can ask at most once.

PR #517, opened 2026-07-24
fix(wiki): header presence matches the mocks' behavior and pixel geometry

nmgarza5:
## Description

Follow-up to #514 against the six header presence mock nodes.

- Panels are mutually exclusive: one `openPanel` union replaces three independent open flags, so the hover card, Auto popover, and +N overflow can never stack.
- Spacing root cause: Opal `Section` emits inline `gap` (default 1rem) and `padding` (default 0) styles that silence gap/padding utility classes. Spacing now flows through Section props and the card measures identical to Figma node 2079:380115 (panel pad 4, title card pad 8 / height 52, avatar inset 8).
- Hover card is the mock's elevated title card + Recent Edits list on the panel surface. Auto panel opens on hover with the Update/Organize rows. Badges are mono Opal marks with 2px identity rings, and the Auto trigger's octagon composite ring tracks the update-warn level via `updateWarnLevel`, shared with UpdatePolicyPanel.
- Removed the inline "Name viewing" line under the doc title, the header stack is the presence surface.
- Review findings applied: shared `OrganizeComingSoonRow`, `patchField` revert arg dropped, badge slots gate on the resolved glyph, comment-header insets restored, `aria-expanded` on panel triggers.

Deferred: update-limit alert panel body (1899-297572), Organize backend field.

## How Has This Been Tested?

Live on dev1: exclusivity probe held exactly one floating panel across both handoff directions, a computed-style probe matches node 2079:380115's values, the hover card renders Recent Edits rows, popover regression and dark mode captured, tsc and prettier through pre-commit on every commit.


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #519, opened 2026-07-25
feat(automanage): selection + reconciliation — the sweep's third step

Bo-Onyx:
Completes the pipeline from the **Wiki Auto Management — Dedup** design page: **detect all → dedup (#516) → rank/select a slate of compatible claims, invalidate the rest.** The sweep becomes a reconciliation: after a full run, the pending set is exactly what the sweep would say today.

## `selection.py`
- **Claims**: a proposal's claim is every path it touches — case-folded (case-insensitive filesystems are a hazard we detect, never manufacture) and subtree-aware (a folder claims everything under it, both directions). Selected proposals have **pairwise-disjoint claims**: at most one live proposal per page as the common case, order-independent approval/execution, and the claim is exactly the agentic applier's sandbox — the selection lock and the execution scope-check are one definition read at two times. `conflicts` is pure (values in, bool out).
- **Cooldown, in its rightful home**: a rejection quiets the `dedup_key`'s content-free prefix (`starts_with`; no stored column) for 30 days — a *new* premise on freshly declined pages is **persisted but unselectable** ("in cooldown until …"). Checked for **REVIVE too**: writing the tests exposed that a cooled row is re-detected every sweep and would slip back in through the revive path after one sweep interval — it now stays at rest until the window passes.

## Runner: collect → reconcile → select
- **Collect** every viable candidate (detect + dedup) before persisting anything — slate composition sees the whole picture instead of emitting as-you-go.
- **Reconcile**: pending rows whose finding neither carried nor revived are invalidated ("not re-detected by run X") — a reviewer is never shown an ask the wiki has outgrown. Full sweeps only (a partial scope's silence proves nothing); **approved rows untouchable** (a human decision in flight is the executor's to re-validate).
- **Select, stability-first**: approved + carried claims hold their pages — a surfaced ask is never bumped, so banners don't churn; new findings fill free pages in registry order (safest-first); losers are persisted invalid ("not selected by run X") so the ledger records **every** detected finding and the row is the revival anchor once its page frees.

**No schema changes** — statuses + reasons carry all of it, per the schema-stability audit on #515.

## Tests (10 new)
Pure claim mechanics (case-insensitive equality, subtree both directions, disjoint pass); one-per-page with registry priority; unselected finding revives once the page frees (same row, `revive_count` 1); pending-not-redetected invalidated; approved untouchable + claim held; carried pending outranks a newcomer even when the newcomer runs first; cooldown persisted-then-quiet-then-revives (including the still-in-window sweep that pins the revive-bypass fix); folder claims lock pages beneath. Full automanage slice: 167 passed.

### DIFF
--- a/backend/app/wiki/automanage/runner.py
+++ b/backend/app/wiki/automanage/runner.py
@@ -222,7 +222,11 @@ def run_detection(
         # Approved rows are a human decision in flight — the executor's to
         # re-validate, never the sweep's to retract.
         invalidated = 0
-        pending_rows = list_by_status(ProposalStatus.PENDING)
+        # limit=None: reconciliation and claim-blocking must see the
+        # COMPLETE live set — a truncated read would leave obsolete
+        # pendings visible and admit claims that conflict with omitted
+        # rows.
+        pending_rows = list_by_status(ProposalStatus.PENDING, limit=None)
         if trigger is TriggerKind.SWEEP:
             revive_ids = {
                 d.existing_id for _, _, d, _ in candidates
@@ -250,7 +254,7 @@ def run_detection(
         # the ledger records every detected finding, and the row is the
         # revival anchor once its page frees up.
         blocked: set[str] = set()
-        for row in list_by_status(ProposalStatus.APPROVED):
+        for row in list_by_status(ProposalStatus.APPROVED, limit=None):
             blocked |= selection.claim_of(row["source_paths"] + row["target_paths"])
         for row in pending_rows:
             if row["id"] in carried_ids:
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
Both addressed in the latest push:

1. **Truncated live sets** — reconciliation and claim-blocking now read pending/approved with `limit=None` (the parameter already existed for exactly this reason on the pending-cleanups list): a default-100 read would leave obsolete pendings visible and admit claims conflicting with omitted rows.
2. **Invalid rows bypassing the cap** — `MAX_PROPOSALS_PER_RUN` now bounds **total ledger writes** per run (selected + persisted-invalid), with the cap log reporting both counts. Deferred drafts re-detect next sweep, so nothing is lost. New test pins it: with a cap of 2 and four contenders for one page, exactly one selected + one invalid persisted, then cutoff.

Bo-Onyx:
@greptile

PR #521, opened 2026-07-27
feat(automanage): card actions backend — dismiss verb + freshness stamp

Bo-Onyx:
Backend for the banner card's final shape, per the pull-only / banner-only design decisions (see the **Wiki Auto Management — Dedup** page). Frontend (buttons + freshness line) stacks next.

## The verb ladder
- **`POST /proposals/{id}/dismiss`** — the missing middle verb (approve / dismiss / reject): `pending → stale` "dismissed by reviewer", **machine-invalidation semantics** — no durable veto, so the finding revives if still (or again) true at a later sweep. Right for "I fixed it myself, clear it now" and "not now". Same writable-envelope gate as approve/reject. Revival now also clears `reviewed_by` — a past dismissal must not read as a human decision on the fresh ask (the auto-apply visibility gate keys on it).

## Freshness ("confirmed by the last scan …")
Transparency over validation: instead of validating proposals at read time (which would eventually tempt LLM calls — now banned by contract), the card shows *when the finding was last verified*. Mechanics: the sweep re-stamps `last_emitted_at` on **carried** pending rows (previously only create/revive stamped it), so the field genuinely means "last asserted against current wiki state"; exposed on `ProposalView`.

Deferred by design (both purely additive later): a `dont_ask_again` flag on reject (page-scoped durable opt-out — the `ai_management_allowed` policy field already exists and is untouched), and a snooze window on dismiss (v1: a still-true dismissed finding returns on the next sweep).

5 tests (dismiss lifecycle incl. revival + reviewer clearing, pending-only guard, carry re-stamps freshness, field exposed to the banner). Automanage slice: 71 passed; ruff + basedpyright clean.

### DIFF
--- a/backend/app/api/automanage.py
+++ b/backend/app/api/automanage.py
@@ -188,13 +188,16 @@ def reject_one(
     every touched page — the caller just proved write on all of them."""
     _require_writable(proposal_id, user)
     proposal = get_proposal(proposal_id)
-    if not review.reject(proposal_id, user_id=user.id):
-        raise HTTPException(status_code=409, detail="proposal is not pending")
+    # Markers before the transition: set_policy is idempotent, so if a write
+    # fails mid-way the proposal is still pending and the call is retryable.
+    # The reverse order would strand partial markers behind a 409.
     if req is not None and req.dont_ask_again and proposal is not None:
         for path in list(proposal["source_paths"]) + list(proposal["target_paths"]):
             update_policy.set_policy(
                 path, ai_management_allowed=False, actor_user_id=user.id
             )
+    if not review.reject(proposal_id, user_id=user.id):
+        raise HTTPException(status_code=409, detail="proposal is not pending")
     return ProposalActionResponse(status="rejected")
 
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
Valid — fixed by inverting the order: the don't-ask markers are stamped **before** the reject transition. `set_policy` is idempotent, so a mid-way failure now leaves the proposal pending and the whole call retryable (a retry re-stamps the same values and completes the reject). Trade-off: a concurrent-actioned proposal can stamp markers and then 409 — acceptable, since the marker was the caller's explicit, independently-authorized intent.

Bo-Onyx:
Moot as of `3e7efb6` — `dont_ask_again` was removed from this PR entirely (deferred; purely additive later), so the reject path no longer writes policy markers at all. When the flag returns, the note here stands: stamp markers only on a successful reject.

Bo-Onyx:
@greptile

PR #527, opened 2026-07-27
feat(automanage): on-create detection rides the nearline queue

Bo-Onyx:
Design alignment (2026-07-27): on-creation work splits in two — **realtime creation context** for agents (separate track, being designed) and **on-creation detection** (this PR). Detection is identical for every attributable creation channel — human UI, API, chat, MCP; only bulk/system channels (ingestion, seeds) skip it, per #525.

One change: `run_detection_on_create` moves offline → nearline. The offline queue can sit behind a whole-space sweep, so a creator could navigate away before their page's finding ever landed. Nearline fits the lane's charter — sub-second, LLM-free, someone is effectively waiting (the creator on the page; an agent mid-conversation). The banner shows the finding during the creation visit; sweeps stay offline.

No schema, no API changes. Existing on-create tests updated to the new lane.


Bo-Onyx:
Closing per design alignment: on-creation splits into (a) realtime creation context for agents — separate track, design first — and (b) on-creation detection, which #525 already ships completely (offline queue stays, per Bo). The write-gate/annotation experiments live in this branch's history if ever needed. Live verification confirmed the #525 dispatch path works end to end (agent write → hook → offline queue → focused run → proposal); earlier ambiguity was a test-harness error, not a product bug.

PR #531, opened 2026-07-27
chore(lint): enforce top-of-file imports (PLC0415)

Bo-Onyx:
Two commits:

1. **Enforce**: ruff `PLC0415` (*import outside top level*) on for `app/` — any new function-local import fails pre-commit/CI unless it carries an explicit `# noqa: PLC0415`. Tests exempt via per-file-ignores (470 sites, their own sweep).
2. **Hoist**: the movable majority of the existing sites now genuinely live in the top import block (read tools' auth imports, seed, metrics, api/wiki draft agents, api/permissions, search, fts, wiki_update, scripts, the leaf-safe half of mcp tools).

What stays lazy keeps its `noqa` **with a comment naming the cycle it breaks**: the mcp `pubsub`/`session`/`tools` triangle, `review → tasks`, `auth → acl`, tools-registry → `wiki_qa` (hoisting this one was tried and produced a live circular import — `tools/__init__` loads `ask_nl_question`, which imports `wiki_qa`), `utils → tasks`, plus optional-dependency lazies (braintrust, boto3, provider SDKs, alembic) and the standalone bootstrap migration.

Verified: ruff + basedpyright clean; the app import graph resolves from both entry directions (`app.main` first and `app.tasks.run_worker` first — the second order is what catches registry cycles).

### DIFF
--- a/backend/app/api/permissions.py
+++ b/backend/app/api/permissions.py
@@ -7,6 +7,8 @@ from typing import Any
 
 from fastapi import APIRouter, Depends, HTTPException, Query, Response, status
 
+from app.db.models import AclEntry
+from app.db.session import session
 from app.auth import User
 from app.auth import groups as groups_repo
 from app.auth import users as users_repo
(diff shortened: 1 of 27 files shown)


Bo-Onyx:
@greptile

PR #537, opened 2026-07-28
feat(images): daily retention sweep, Prometheus metrics, backup sizing note

nmgarza5:
## Summary

Retention sweep and observability for stored wiki images. Stacks on the image store/endpoints PR.

- Daily sweep (`app/tasks/image_sweep.py`, 12:00 UTC on `lightweight_maintenance_queue`, mirroring the ingest-eval retention prune): one batched `git grep -F` over the working tree including `.trash/` (a documented exception to the `list_paths` exclusion, via a new wrapper in `app/wiki/git.py`) plus the live coedit session buffers. Referenced images get their flag cleared. Unreferenced images past a 24h creation grace get flagged (`unreferenced_since`, new nullable column, migration `e7a1c4d9b2f6`), and flagged images past the trash retention window (`TRASH_RETENTION_DAYS`, 30 days, 0 disables) are deleted. Unparseable timestamps never delete.
- One deliberate model choice for review: never-referenced uploads are flagged at 24h and deleted after the retention window rather than fast-deleted at 24h, because a snapshot scan cannot distinguish never-referenced from just-dereferenced without extra schema or git-history digging that would break the lightweight-queue contract. Abandoned uploads linger ~31 days; the storage cost of that is trivial.
- Prometheus: `wiki_images_total` and `wiki_images_bytes_total` gauges (refreshed each sweep, even on failure), `wiki_image_sweep_deleted_total`, and `wiki_image_upload_rejected_total{reason}` on the upload endpoint's 413/415 paths.
- `docs/backups.md`: images live in Postgres and already ride the pg_dump; dump size now grows with stored image bytes and pre-compressed formats do not shrink, so size the backup CronJob's ephemeral storage accordingly.

## How Has This Been Tested?

`tests/test_wiki_image_sweep.py`, 8 tests against real Postgres + real git: committed-page reference spared, live coedit buffer reference spared, `.trash/`-only reference spared (end-to-end through a real trashed page), flag set then cleared when a reference returns, deletion after the retention window (timestamps manipulated), creation-grace skip, metrics movement, upload-rejection counter. Combined with the 14 image endpoint tests: 22 passed. ruff and strict basedpyright clean.

### DIFF
--- a/backend/app/api/images.py
+++ b/backend/app/api/images.py
@@ -75,7 +75,7 @@ async def upload_image(
         await run_in_threadpool(image_store.delete, image_id)
         raise HTTPException(status_code=404, detail="anchor page not found")
 
-    url = f"/api/wiki/images/{image_id}"
+    url = image_store.serving_url(image_id)
     alt = (filename or "").replace("\n", " ").replace("\r", " ").replace("]", "")
     return UploadImageResponse(id=image_id, url=url, markdown=f"![{alt}]({url})")
 
--- a/backend/app/api/images.py
+++ b/backend/app/api/images.py
@@ -7,6 +7,7 @@ from starlette.concurrency import run_in_threadpool
 
 from app.auth import User, require_can
 from app.auth.deps import require_user, require_user_or_bearer
+from app.metrics import wiki_image_upload_rejected_total
 from app.models.images import UploadImageResponse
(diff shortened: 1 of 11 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

PR #546, opened 2026-07-29
fix(wiki): Auto popover toggles in place on a cache the panel shares

nmgarza5:
## Description

**Bug.** Toggling a switch in the Auto popover yanked you into the side panel, and the two surfaces disagreed: each fetched and PATCHed `/update-policy` on its own, so a write on one left the other stale. The popover's sub-toggle switches also overhang its right edge, because the indent is a left margin that shifts the whole row and pushes the right-aligned switches out of bounds.

**Fix.**
- Popover and panel share one SWR key (`useUpdatePolicy`) and one writer (`saveUpdatePolicy`, optimistic with `rollbackOnError`).
- Popover switches toggle in place and the popover stays open. The panel opens only from the Auto mark, or the Page Instructions expander, which opens it straight into the editor.
- Indent becomes a padding wrapper so the row's right edge stays pinned, matching the group `UpdatePolicyPanel` already renders.
- `keepPreviousData: false` on the policy hook. It is on app-wide, and the switches gate on a loaded policy rather than `isLoading`, so one scope's values could render and PATCH against another path.
- Removes the now-dead `getUpdatePolicy`.

The Organize toggle from #541 is unchanged.

## How Has This Been Tested?

| Before | After |
|---|---|
| before | after |

Both switches sat 11px past the panel's right edge and now sit 13px inside it, aligned with each other.

Captured in Chrome against the real Opal stylesheet at the real panel width (`--block-width-panel-medium-small`, 22.5rem), rendering the actual Opal primitives with each indent variant. The measurements match the running app exactly (switch right edge 395 vs panel right 384 before, 371 vs 384 after).

`bun run typecheck` and prettier are clean. No Python in the diff.

### DIFF
--- a/frontend/src/components/wiki/PresenceAvatars.tsx
+++ b/frontend/src/components/wiki/PresenceAvatars.tsx
@@ -22,6 +22,7 @@ import {
   AnchoredPanel,
   AutoGlyph,
   PolicyPopover,
+  type OpenUpdatesPanelOpts,
 } from "@/components/wiki/policyPanels";
 import { relativeTime } from "@/lib/users";
 import { fetchFileHistory } from "@/lib/wiki/svc";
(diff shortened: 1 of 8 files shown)


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #548, opened 2026-07-29
feat(wiki): AI Auto-Edits master toggle (frontend)

Bo-Onyx:
Stacked on #547 (the `ai_edits_disabled` backend). Implements the mocks + the Slack spec:

- **AI Auto-Edits is a switch again**, bound to the master field. It **overrides without changing** Update/Organize — their stored values are preserved server-side and this panel receives the display view, so re-enabling brings them back exactly as they were.
- **Children show/hide with the master** (hidden while off), in both the side panel and the header popover.
- **The limit bar hides while the master is off** (no auto-edits to limit); the limit itself is server state and is not reset by toggling. **Update History stays visible regardless** — it is the page's full edit record.
- Side panel master switch carries the same inherit-origin tooltip + hover reset affordance as the other rows; the popover writes explicit booleans as it does for the children.
Screenshot 2026-07-29 at 10 58 09 AM
Screenshot 2026-07-29 at 10 58 01 AM

`bun run typecheck` + prettier clean.


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #549, opened 2026-07-29
Use stock postgres/nginx images; drop pg_textsearch + pgmq

Bo-Onyx:
## What & why

The custom `agent-wiki-postgres` image existed only to bake in two Postgres extensions — `pg_textsearch` (BM25) and `pgmq` (queue). **Both were migrated away:** BM25 search now runs on OpenSearch (`app/db/fts.py`, `opensearch-py`) and task queues on Redis Streams (`app/tasks/queue.py`, `redis-py`). No app code, migration, or runtime config touches either extension anymore — the migrations `0007` (pgmq→Redis Streams) and `0012`/`0013` (pg_textsearch→OpenSearch) are explicit no-ops, and there are zero live `CREATE EXTENSION` / `shared_preload` / raw-SQL sites left.

So the custom image and its `shared_preload_libraries=pg_textsearch` boot flag were dead weight. This points both services at **stock public images**:

- **postgres** → `postgres:17`
- **nginx** → `nginx:1.27-alpine` with `nginx.conf` bind-mounted

### Two nice consequences

1. **Advances #545's goal.** #545 wanted a lighter pull-based Compose install (8 GB → ~2.5 GB); the postgres image was the expensive part (it compiled two C extensions from source). Stock images are plain pulls, no compile.
2. **Removes the Docker Hub blocker.** The nightly build was failing because `onyxdotapp/agent-wiki-{nginx,postgres}` don't exist on Docker Hub and the scoped `DOCKERHUB_TOKEN` can't create them (`401 ... insufficient scopes`). With nothing custom to publish for these two, that failure goes away — no admin/repo-provisioning needed.

## Changes

- `docker-compose.yml` — stock images; drop the `build:` contexts, the `shared_preload_libraries=pg_textsearch` command, and bind-mount `nginx.conf`
- delete `deploy/postgres/Dockerfile` and `nginx/Dockerfile`
- drop `postgres` + `nginx` from the nightly build matrix and the `v*` release workflow
- update `deploy/README.md`, `deploy/terraform/README.md`, `CLAUDE.md` — stale extension / raw-SQL references

## Scope / safety

- **Helm/prod unaffected** — the chart references neither custom image (it uses `ingress.yaml`, not an nginx pod, and expects an external stock Postgres). This is confined to the Compose install path, exactly what #545 touched.
- **Install path safe** — `install.sh` does a full `git clone`, so `nginx/nginx.conf` is always present for the bind-mount.
- `docker compose --profile app config` validates; both workflow YAMLs parse.
- On-disk data compatible (custom image was `FROM postgres:17`). Any pre-existing dev DB that once ran `CREATE EXTENSION` will have inert leftover objects, harmless with the preload flag gone.

Effectively supersedes the postgres/nginx publishing half of #545.


Bo-Onyx:
@greptile

PR #550, opened 2026-07-29
fix(coedit): recover saves across backend workers

nmgarza5:
## Description

A disconnect handled by one backend worker could close a co-edit session while another worker still had an active socket, causing later saves to fail until reload.

This stops treating one process's socket registry as shared presence. The existing `coedit_participants` heartbeat now owns presence expiry, session closure waits for an empty roster and clean buffer, and the client reconnects and reapplies its local document if the server reports an inactive session. No schema or migration changes.

## How Has This Been Tested?

- Full backend test suite
- 108 focused co-edit and migration tests
- PR-range pre-commit checks: Ruff, BasedPyright, TypeScript, Prettier, whitespace, and secret scanning
- Independent full-diff Codex review

## Additional Options


raunakab:
"This PR should not merge until co-edit leases and leave tasks remain safe while old and new backend workers coexist during a rolling deployment." - Greptile.

@nmgarza5 seems like this specifically is not applicable to use right now.

@greptile-apps, please re-review, but keep in mind that rolling deployments are not an issue for us just yet.

backend/app/api/coedit.py
```diff
@@ -370,23 +367,4 @@ async def ws(websocket: WebSocket, path: str, user: User = Depends(require_user_
...
-            await asyncio.to_thread(record_leave, sess.id, user.id)
-        except asyncio.CancelledError:
-            try:
-                leave_coedit_session(sess.id, user.id)  # enqueue, no await
-            except Exception:
-                log.exception(
-                    "coedit: queued-leave fallback failed for session %s", sess.id
-                )
-            raise
         log.info("coedit ws closed session=%s user=%s", sess.id, user.id)
```
Bo-Onyx, line 370: The immediate commit on disconnect is gone.

`record_leave` ended with `checkpoint_coedit_session(...)`, so an abruptly-dropped tab reached git in ~1s. Now `finally` only calls `disconnect()`, so the commit waits on the presence scan — 60s stale + up to 60s scan. Normal teardown is fine (`hooks.ts` checkpoints before closing), but force-quit / crash / network drop / pod restart don't run it. Nothing is lost, but readers of committed content (other viewers, search, `read_doc`, agents) see stale text for up to 2 minutes.

Restore just the checkpoint, not the leave:

```python
    finally:
        if conn is not None:
            coedit_channel.disconnect(conn.id)
        try:
            checkpoint_coedit_session(sess.id)
        except Exception:
            log.exception("coedit: checkpoint enqueue failed on disconnect for session %s", sess.id)
        log.info("coedit ws closed session=%s user=%s", sess.id, user.id)
```

This can't reintroduce the bug you're fixing — that was the local `user_still_connected` guess plus the row delete after it. The checkpoint asks no liveness question, reads only DB state, no-ops on a clean session, and its close is gated by your new `NOT EXISTS(participants)` predicate. Already imported for `_handle_checkpoint`.

PR #551, opened 2026-07-30
fix(coedit): shared presence heartbeat instead of a per-process socket registry

Bo-Onyx:
Continues #550 (closed) — Nik's two commits are preserved here, plus a fix for a regression they introduced and one follow-up.

## The bug

`coedit_channel`'s connection registry is per-process, but the backend runs `uvicorn --workers 2`, so a socket closing on one worker could not see the same user's live sockets on the other. `record_leave` asked `user_still_connected` anyway, concluded the user was gone, deleted their `(session_id, user_id)` participant row, found the roster empty, and closed the shared session. Every later message from the still-open sockets then failed with `no_active_session` until the page was reloaded.

Reproduced before the fix — one user, four tabs, closing **one**:

```
tab0..tab3: joined session=52 version=0
before: op_result ok=True  version=1
--- closing tab0 only (3 tabs still live) ---
after:  op_result ok=False error=no_active_session
```

The session row was left `closed` with three healthy sockets attached. Note it needs no second person: two tabs, or one tab plus an overlapping reconnect, is enough.

## The fix (Nik's commits)

Liveness moves from the in-process registry to the participant row's heartbeat, which any live connection refreshes from any process:

- `user_still_connected` and `_user_of` deleted; `coedit_channel.connect()` no longer takes a `user_id`.
- `coedit_leave.py` deleted. Its queued fallback ran on a worker, where the channel registry is always empty, so its still-connected guard always answered "gone" — the same flaw a single-worker mitigation would not have covered.
- The `_send_loop` heartbeat is wall-clock instead of idle-only, so a connection carrying a steady broadcast stream still refreshes its row.
- `expire_stale_participants` removes rows past four missed heartbeats from the periodic scan, broadcasts presence for affected sessions, and checkpoints the ones that emptied.
- `join()` returns `False` under a session row lock with a retry in `_connect_sync`, and `close_if_clean` carries its own participant predicate, so a join can't land in a closing session.
- `touch()` returning `False` ends the send loop, so an expired connection closes and the client's reconnect replays its local document.

No new table and no migration: presence is per-user, and `last_seen_at` already carried cross-process liveness.

## Regression fixed here

The immediate checkpoint on disconnect lived inside `record_leave`, so deleting that function took it along. Normal teardown still commits (the client checkpoints before closing), but force-quit, crash, network drop, and pod restart don't run that path — those edits waited on the presence scan instead, up to the stale window plus a scan interval. Nothing was lost, but readers of committed content (other viewers, search, `read_doc`, agents) served stale text for that window. `test_disconnect_of_last_participant_checkpoints` was removed with the module and is restored here as `test_disconnect_commits_buffer`.

Restoring the checkpoint alone can't reintroduce the bug: it asks no liveness question, reads only shared state, no-ops on a clean session, and its close is gated by `close_if_clean`'s participant predicate.

## Follow-up included

`close_if_clean` only ever runs for a session whose last participant just expired, so a session that empties by any other route stays `active` forever — and being clean, the checkpoint scan skips it too. It keeps holding the active-path unique index, so page moves are refused, and every new viewer adopts its buffer instead of reading HEAD. We have one in production dating from Jul 23. `close_abandoned_sessions` applies the same predicate set-wise from the periodic scan.

## Not in scope

Two independent problems, each worth its own PR:

- The checkpoint write-back CAS loses to any op landing mid-commit, and that branch leaves both `checkpointed_version` and `base_sha` unadvanced — so the session stays permanently dirty, the scan re-enqueues forever, and every later autosave 3-way merges against an ever-staler base. In production this produced ~124 checkpoint commits in a few hours on one page, a commit re-inserting two lines that had existed for 75 minutes, and another splicing text mid-word.
- `resync` is unactionable client-side: a buffer rebase writes no `coedit_ops` row, so `pull()` receives an empty op list, ignores `current_head_version`, and the client silently keeps a stale document — after which its ops anchor against text the server doesn't have.

## Testing

- 98 co-edit tests pass (`test_coedit_api`, `test_coedit_repo`, `test_coedit_checkpoint`, `test_coedit_channel`), including the restored disconnect-commit test and a new sweep test.
- ruff and basedpyright clean.


Bo-Onyx:
@greptile

PR #553, opened 2026-07-30
fix(coedit): stop holding a thread per open socket

Bo-Onyx:
Third of the co-edit fixes, after #551 and #552. Independent of both.

## The problem

`asyncio.to_thread` uses the event loop's **default executor**, whose default size is `min(32, cpu + 4)`. Measured in the running pod: `cpu_count` is 2, so **6 threads**.

A co-edit WebSocket occupies one of those for its entire life. `_send_loop` blocks in `coedit_channel.drain` (a `queue.Queue.get(timeout=1.0)`) and re-enters it the moment it returns, so at steady state one thread per open socket is parked. That leaves a web process able to hold roughly five concurrent editors; beyond that every offloaded call in the process — the connect handshake, every op, cursor ping and checkpoint, for every session — queues behind a free thread.

Worth noting the old SSE transport didn't have this ceiling: Starlette runs sync routes through anyio's thread limiter, which defaults to **40** tokens. Verified both numbers in the pod:

```
anyio default thread limiter (sync routes): 40
asyncio default executor (to_thread):       6
```

The module comment on `_SEND_LOOP_POLL_SECONDS` already describes what saturation looks like — "reads as 'the whole app froze'… a self-reinforcing spiral" — and notes it was hit in production.

## The fix

Install an explicit executor in the lifespan, sized by a new `WEB_THREAD_POOL_SIZE` (default 64). This is a connection ceiling, not a CPU knob, which is why it isn't derived from core count.

## Measured

Same backend, same probe, only the knob changed — N sockets open, then time an op on the last one:

| `WEB_THREAD_POOL_SIZE` | sockets | op latency |
|---|---|---|
| 6 (the old default) | 10 | 1.02 s |
| 6 | 20 | 3.02 s |
| 64 | 20 | 0.01 s |
| 64 | 40 | 0.01 s |

The stalls land on whole multiples of the 1 s drain poll, which is the queueing behind parked drain threads.

## Follow-up, deliberately not here

The structural fix is for the send loop to stop occupying a thread at all: an `asyncio.Queue` fed from `_deliver_local` via `call_soon_threadsafe`, instead of a `queue.Queue` drained in a thread. That removes the per-connection thread entirely, but it reworks the `CLOSE_SIGNAL` / `wake` handshake that carries a fair amount of hard-won reasoning, so it doesn't belong in the same change as raising the ceiling.

## Testing

2056 passed, 2 skipped. `conftest.py`'s `Config(...)` needed the new field. ruff and basedpyright clean.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -29,6 +29,11 @@ DATABASE_URL=postgresql://postgres:postgres@localhost:5432/agent_wiki
 # at this size. Default 1000; applies to all three pgmq queues.
 MAX_QUEUE_SIZE=1000
 
+# Threads the web process offloads blocking work onto (`asyncio.to_thread`).
+# Every open co-edit WebSocket holds one for its send loop, so this is the
+# ceiling on concurrent editors per process, not a CPU knob. Default 64.
+WEB_THREAD_POOL_SIZE=64
+
 # Browser-facing origin the backend advertises (launcher URIs, redirect
 # targets, email links, MCP tool-result URLs) — REQUIRED, never inferred
 # from request headers.
--- a/.env.example
+++ b/.env.example
@@ -33,11 +33,6 @@ MAX_QUEUE_SIZE=1000
 # git calls) via `asyncio.to_thread`. Default 64.
 WEB_THREAD_POOL_SIZE=64
 
-# Threads for the co-edit send loops — one held per open socket for its whole
-# life, so this is the ceiling on concurrent editors per process. Kept separate
-# from the pool above so sockets can't starve request work. Default 256.
-COEDIT_SOCKET_POOL_SIZE=256
-
 # Browser-facing origin the backend advertises (launcher URIs, redirect
 # targets, email links, MCP tool-result URLs) — REQUIRED, never inferred
 # from request headers.
--- a/.env.example
+++ b/.env.example
@@ -29,10 +29,14 @@ DATABASE_URL=postgresql://postgres:postgres@localhost:5432/agent_wiki
 # at this size. Default 1000; applies to all three pgmq queues.
 MAX_QUEUE_SIZE=1000
 
-# Threads the web process offloads blocking work onto (`asyncio.to_thread`).
-# Every open co-edit WebSocket holds one for its send loop, so this is the
-# ceiling on concurrent editors per process, not a CPU knob. Default 64.
+# Threads the web process offloads short-lived blocking work onto (DB reads,
+# git calls) via `asyncio.to_thread`. Default 64.
 WEB_THREAD_POOL_SIZE=64
+
+# Threads for the co-edit send loops — one held per open socket for its whole
+# life, so this is the ceiling on concurrent editors per process. Kept separate
+# from the pool above so sockets can't starve request work. Default 256.
+COEDIT_SOCKET_POOL_SIZE=256
 
 # Browser-facing origin the backend advertises (launcher URIs, redirect
 # targets, email links, MCP tool-result URLs) — REQUIRED, never inferred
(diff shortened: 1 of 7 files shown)


Bo-Onyx:
Valid, and it found a design flaw rather than a doc nit — pushed `27cf73e6`.

Raising the shared pool's size only moved the cliff. A drain thread is held for the connection's whole life, so socket occupancy and the short-lived offloads were competing for the same threads: at the ceiling every thread is parked in a drain and the handshakes, ops, cursor pings, checkpoints and heartbeats stall behind them. My "~64 editors with headroom for transient borrows" claim above was wrong — at 64 sockets there is no headroom.

The drains now have their own executor (`COEDIT_SOCKET_POOL_SIZE`, default 256), so the two can't compete. `WEB_THREAD_POOL_SIZE` bounds only work concurrency now, which is the thing its size should be reasoned about.

Re-measured with the work pool pinned at the old default of 6:

| work pool | drains | sockets | op latency |
|---|---|---|---|
| 6 | shared | 10 | 1.02 s |
| 6 | shared | 20 | 3.02 s |
| 6 | **separate** | 10 | 0.01 s |
| 6 | **separate** | 20 | 0.01 s |
| 6 | **separate** | 40 | 0.01 s |

Six work threads now serve 40 sockets with no measurable delay, because sockets no longer touch that pool.

A connection ceiling still exists — socket 257 gets no send loop — but it's now localized to the excess connection instead of degrading every session in the process. Removing it entirely is the `asyncio.Queue` + `call_soon_threadsafe` change noted in the description, which drops the per-socket thread altogether.

2056 passed, 2 skipped; ruff and basedpyright clean.

Bo-Onyx:
@greptile

Bo-Onyx:
Valid, and it's the same objection as the last one a level down: any fixed number leaves a silently half-dead socket. So I dropped the number — pushed `3bc3e091`, which removes the per-socket thread instead.

The thread only existed because the send loop read its queue with a blocking `queue.Queue.get`. A connection now carries a `notify` callback: the publisher calls it after enqueueing, from whatever thread it's on, and it schedules an `Event.set` onto the loop. The send loop awaits that event, so nothing is held on a connection's behalf and there's no per-connection ceiling to exceed.

Verified live — **300 concurrent sockets** served by a work pool of **6**, ops at 0.02s, **15 threads** in the whole process:

```
work pool=6, no drain threads, 20 sockets   → op_result after 0.02s
work pool=6, no drain threads, 100 sockets  → op_result after 0.01s
work pool=6, no drain threads, 300 sockets  → op_result after 0.02s
```

`COEDIT_SOCKET_POOL_SIZE` and the dedicated drain pool are gone again; `WEB_THREAD_POOL_SIZE` stays, bounding only short-lived DB/git offloads.

Two things worth flagging from doing it:

- **`Connection.post` is now the only way to enqueue.** A bare `put_nowait` lands the frame but leaves the loop asleep until its next heartbeat — which is exactly what my first draft did, turning every op and checkpoint into a 15-second round trip (`test_get_ops_returns_missed_changes_for_rebase` went from 0.2s to 45s, which is how I caught it). The channel tests now assert the wakeup, not just delivery, so that can't regress silently.
- **`CLOSE_SIGNAL`, `wake`, and the 1s poll all go away.** Cancelling the send task interrupts an `Event.wait` directly, so teardown no longer needs a sentinel pushed into a connection's own queue to unblock a thread that couldn't be cancelled.

2056 passed, 2 skipped; ruff and basedpyright clean. The two functional probes still pass: closing one of several tabs keeps the session alive, and an abrupt TCP abort commits to git in ~1s.

Bo-Onyx:
@greptile

PR #554, opened 2026-07-30
feat(coedit): get_session — the catch-up a resync actually needs

Bo-Onyx:
Backend half of the last co-edit correctness item, after #551, #552 and #553. Frontend stacked on top.

## The gap

`resync` tells participants the buffer was **replaced** out of band — a checkpoint's merge folded in a committed change, or an inbound agent commit was rebased into the session. `rebase_onto` writes no `coedit_ops` row (deliberately: it isn't a co-edit op), and `get_ops` is the only read the WS protocol offers.

So a client handling a resync calls `get_ops`, receives an empty list, and keeps its pre-merge document — permanently. From then on its ops are anchored against text the server doesn't have, and `apply_op` only bounds-checks, so they splice in at the wrong offsets. That's the "different people see different things" half of the original report, and the mechanism behind the production commit that left `# Bugsemove active agent bar from top of wiki doc` with a stray `- [ ] r`.

The gap predates the WebSocket migration: the old transport had `GET /coedit/session`, but `svc.ts` never called it on resync either.

## The change

`get_session` returns the buffer plus the version it is current as of, so a caller can reset its document and its collab baseline together. Read-gated and `request_id`-correlated like the other request frames.

The two catch-up reads now cover the two ways a client falls behind, which the module docstring spells out:

- `get_ops` — a missed frame or a stale-version reject. Replays the op log, preserving unconfirmed local edits.
- `get_session` — a resync. Re-reads the buffer, because there is no op to replay.

## Also here

The two rejections that were silent are now logged: `invalid_op` at warning with the offending offsets (it means the sender's document has diverged — the shape a desync takes), and `stale_version` at debug (ordinary under concurrent editing, but a storm of them should be visible). Their absence is why none of this appeared in the logs while it was happening.

## Tests

- `test_get_session_returns_the_live_buffer` — replaces the buffer through `rebase_onto` the way a checkpoint merge does, then asserts `get_session` returns the merged text at the bumped version **and** that `get_ops` returns `[]` for the same interval. That second assertion is the point: it pins down why this endpoint has to exist.
- `test_get_session_on_closed_session_reports_terminal_error` — `no_active_session` rather than a hanging promise.
- `test_get_session_requires_read` — succeeds, then the read grant is revoked mid-session and it returns `forbidden`.

2059 passed, 2 skipped; ruff and basedpyright clean.

## Next

The frontend change consumes this: on `resync`, fetch the session and reset the editor's document and collab version instead of calling `pull()`. There's a design call to make there — unconfirmed local edits can't be rebased across an out-of-band replacement, so they're either dropped or flushed first — which is worth deciding explicitly rather than in passing.

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -23,6 +23,13 @@ vs viewers client-side from the live caret frames — a rendered caret IS the
 Participants leave after their shared heartbeat expires. A disconnect never
 deletes shared presence based on one process's local socket registry — it
 only commits the buffer, which needs no liveness information.
+
+Two catch-up reads, for the two ways a client can fall behind. ``get_ops``
+replays the op log after a missed frame or a stale-version reject, preserving
+the client's unconfirmed edits. ``get_session`` re-reads the whole buffer, for
+a ``resync`` — the buffer was *replaced* by a git commit folded into the
+session (a checkpoint merge, an inbound agent edit), which logs no op and so
+cannot be replayed.
 """
 
 from __future__ import annotations
@@ -43,12 +50,14 @@ from app.models.coedit import (
     CheckpointResultFrame,
     CursorMessage,
     GetOpsMessage,
+    GetSessionMessage,
     JoinedFrame,
     OpMessage,
     Operation,
     OpResultFrame,
     OpsResultFrame,
     ParticipantOut,
+    SessionResultFrame,
 )
 from app.tasks.coedit_checkpoint import checkpoint_coedit_session
 from app.wiki import coedit, coedit_channel, git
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
Valid — fixed in `f29cc08d`. Dumping each `Change` whole put the `insert` (arbitrary page content, arbitrary length) into production logs, which I shouldn't have done.

Logging offsets plus `insert_len` rather than dropping the insert entirely — the length distinguishes a stray keystroke from a large paste when reading a desync, and costs nothing:

```python
[{"from": c.from_, "to": c.to, "insert_len": len(c.insert)} for c in msg.changes]
```

30 WS tests pass; ruff and basedpyright clean.

Bo-Onyx:
@greptile

Bo-Onyx:
Folding this into #555 — the stacked branch already carries these commits, and reviewing the protocol addition next to the client that consumes it is easier than reviewing it alone. No code changes; the commits are identical.

PR #555, opened 2026-07-30
fix(coedit): resync — re-read the buffer the op log can't replay

Bo-Onyx:
Was #554 (backend) + a stacked frontend PR; folded into one since the protocol addition and the client that consumes it are easiest to review together. Last of the co-edit fixes after #551, #552, #553.

## The bug

`resync` announces the buffer was **replaced** out of band — a checkpoint's merge folded in a committed change, or an agent/ingest commit was rebased into the open session (`rebase_onto`). That replacement is a git commit, not a co-edit op, so it writes no `coedit_ops` row. The client's handler was `pull()`, i.e. `get_ops` — which returned `[]`, early-returned, and never looked at `current_head_version`.

So the client kept the pre-merge document at its old version, permanently:

- every push carried a stale `base_version` → `stale_version` → pull → `[]` → repeat, no backoff — the "Couldn't save" people saw
- if versions ever re-aligned, ops applied at offsets that meant something else in the server's buffer — the production commit that left `# Bugsemove active agent bar from top of wiki doc` with a stray `- [ ] r`

Only a reload recovered. The gap predates the WS migration (the old `GET /coedit/session` existed but was never called on resync); #552 was what made it fire constantly.

## Backend

- **`get_session`** WS message: the buffer plus the version it is current as of, read-gated, `request_id`-correlated. The key test replaces the buffer via `rebase_onto` exactly as a checkpoint merge does, then asserts `get_session` returns the merged text **and** `get_ops` returns `[]` over the same interval — pinning down why the endpoint must exist.
- The two silent rejections now log: `invalid_op` at warning (offsets + insert *length* only — the text is page content and stays out of logs), `stale_version` at debug.

## Frontend

**Cheap cause first.** `broadcast_op` degrades to a resync when an op exceeds the bus payload cap; that op *is* in the log, so `pull()` closes the gap with pending edits untouched. Only if the log can't reach the announced version was the buffer replaced — then `get_session`.

**The replacement is handed to collab as the update the server never wrote**: one synthesized change from the last-confirmed text to the server's, padded with empty updates when several versions passed. `receiveUpdates` applies it, advances the version, and rebases un-acked local edits over it — so typing survives (an edit inside the rewritten span lands at its boundary, visible and fixable, rather than silently dropped), and caret/selection/undo survive because the editor is never rebuilt.

**Invariant to watch in review:** the synthesized change needs the last-*confirmed* document, which is unrecoverable from current state (inverting un-acked ChangeSets needs the document they applied to). It is tracked, and every confirmation routes through one `confirmUpdates` helper. A path that bypasses it diffs against the wrong base and mis-anchors edits — the bug class this closes.

Also: `doPush` backs off exponentially on rejections a rebase can't clear (`invalid_op` used to re-push forever), and autosave moves 2s/30s → 15s/3min (the checkpoint engine is one-commit-per-session by design; durability rides on Postgres + teardown/`pagehide` checkpoints).

## Testing

- Backend: full suite green; new `get_session` tests (live-buffer read, closed-session error, mid-session read revocation).
- Collab semantics headlessly (`EditorState` needs no DOM): 3 un-acked keystrokes + a v7 replacement → merged line **and** all three keystrokes kept, still queued (`"Xalpha\nbeta\ngamma\nYZW"`); version padding correct; in-span edit kept at the boundary (`"hello universeWORLD\n"`).
- End-to-end in a real browser against a 2-worker backend running this branch: typed live, forced a buffer replacement server-side, watched the editor converge on the merged text with local typing intact — full transcript in the PR thread.

No committed frontend test: there's no test runner in `frontend/` today. Worth a follow-up (the `confirmUpdates` invariant deserves CI), but that's runner + CI wiring, not this PR.

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -23,6 +23,13 @@ vs viewers client-side from the live caret frames — a rendered caret IS the
 Participants leave after their shared heartbeat expires. A disconnect never
 deletes shared presence based on one process's local socket registry — it
 only commits the buffer, which needs no liveness information.
+
+Two catch-up reads, for the two ways a client can fall behind. ``get_ops``
+replays the op log after a missed frame or a stale-version reject, preserving
+the client's unconfirmed edits. ``get_session`` re-reads the whole buffer, for
+a ``resync`` — the buffer was *replaced* by a git commit folded into the
+session (a checkpoint merge, an inbound agent edit), which logs no op and so
+cannot be replayed.
 """
 
 from __future__ import annotations
@@ -43,12 +50,14 @@ from app.models.coedit import (
     CheckpointResultFrame,
     CursorMessage,
     GetOpsMessage,
+    GetSessionMessage,
     JoinedFrame,
     OpMessage,
     Operation,
     OpResultFrame,
     OpsResultFrame,
     ParticipantOut,
+    SessionResultFrame,
 )
 from app.tasks.coedit_checkpoint import checkpoint_coedit_session
 from app.wiki import coedit, coedit_channel, git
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
End-to-end verification, run against this exact branch — both images built from it and swapped into the compose stack (2 uvicorn workers, nginx, real Postgres/Redis bus), driven in a real Chrome session:

1. Opened a page, typed ` TYPED-BEFORE-RESYNC` in the live editor (acked, session at v20).
2. Replaced the buffer server-side the way a live-rebase does — `coedit.rebase_onto` appending `AGENT-MERGED-LINE` (v21, no op row) + `broadcast_resync` from a separate process, so the frame traveled the Postgres NOTIFY bus.
3. **The merged line appeared in the editor with no reload; the typed text and the caret survived.**
4. Typed ` TYPED-AFTER-RESYNC` — accepted, session advanced v21 → v40, and it landed at the caret, i.e. at offsets correct in the *merged* buffer:

```
[before]PEER-EDIT hello from probe BROWSERTEST XYZ TYPED-BEFORE-RESYNC TYPED-AFTER-RESYNC

AGENT-MERGED-LINE
```

5. The 15s autosave then checkpointed exactly that content to git.

On main, step 3 shows nothing (pull returns `[]` and the client stays pinned at v20), step 4 is an endless `stale_version` loop — the "Couldn't save" from the meeting — and only a reload recovers.

Also ran the full backend suite on the combined branch: 2059 passed, 2 skipped.

Bo-Onyx:
@greptile

PR #557, opened 2026-07-30
feat(editor): Tiptap/Yjs live editor — frontend (on the room-less backend)

Bo-Onyx:
Stacked on #556. **Base is `bo/yjs-takeover`, not `main`** — review and merge #556 first.

Recreates #542's frontend on top of the room-less backend. #542 is based on `feat/editor`, whose backend #556 replaces wholesale, so retargeting it would have dragged the old backend along; this takes the frontend only.

## How it was ported

File by file, **not** by checking out #542's frontend tree — `main`'s frontend moved on since their common base, and a tree checkout is exactly what silently reverted `main`'s work in the two earlier split attempts.

Six files both sides touched:

- `lib/editor/{svc,hooks,components,constants}` — the OT co-edit client this replaces. `main`'s changes there were protocol fixes to a client that no longer exists, so #542's versions win.
- `views/wiki/FileView.tsx`, `components/wiki/PresenceAvatars.tsx` — real three-way merge. `main`'s `OpenUpdatesPanelOpts` / edit-instructions deep-link plumbing is preserved; verified after the fact that it's still wired.

No `get_ops` / `base_version` / `stale_version` remains anywhere in `src/`.

## Adapted to the room-less protocol

**Dropped the whole `resync` path** — `onResync`, `disconnectForResync`, the handler branch. The server has no such frame any more, and this is a user-visible improvement rather than cleanup: #542 responded to a checkpoint merge or an agent's out-of-band commit by closing the socket and reconnecting, which discarded the caret. The same content now arrives as an ordinary Yjs update integrated in place, so the caret survives and pending local edits rebase over it.

**`checkpoint_result` reason surfaced** — the frame carries `forbidden` (a viewer trying to save) or the task's own failure, instead of a generic "checkpoint failed".

## Bugs fixed on top of #542

**Reconnects clobbered active sockets** (review). `sockets` was keyed by session id, but a session outlives any single socket — a reconnect joins the *same* session id while the previous socket may still be closing. The replacement overwrote the old entry, the old socket's `onclose` then deleted the replacement's, and `closeSession`/`checkpointSession` resolved to whichever socket happened to be registered. The concrete damage: the effect teardown doesn't await its own async close, so React starts the next effect body first; the teardown's later `closeSession(sid)` then closed the *freshly established* socket and marked it an **expected** close, which tells the reconnect loop to give up. Silently dead editor, and saves failing with "not connected". Now keyed per connection, with the handle exposed on `CoeditConnection` so the hook acts on the socket it owns.

**Comment rail only shifted the title.** `globals.css` reserved the 360px lane with `.rail-reserved .cm-scroller .cm-content` — CodeMirror's DOM. Tiptap renders `.editor-prose > .ProseMirror`, so the editor text never reserved the lane while `.rail-inset` (title, banners) did, and the two fell visibly out of alignment. Same dead selector cost the editor's scrollbar-hiding, so it showed native scroll chrome next to `EditorEdgeScrollbar`'s tick rail, which is still the intended indicator and still rendered.

**Placeholder appeared mid-document.** Pressing Enter part-way down a written page put "Start typing, or pick a template above…" in the middle of it — wrong twice over, since there is nothing to start and a template can't be applied to a page that already has content. The rule was keyed on `.is-empty`, which the Placeholder extension applies to any empty textblock holding the cursor; `.is-editor-empty` lands on the same decoration but only when `editor.isEmpty`. Checked against the installed extension's source, not the docs.

**`[text](url)` didn't become a link.** StarterKit ships the `link` mark and autolinks a bare URL, but nothing converted markdown link *syntax* — the first link form anyone writing markdown reaches for. Safe to author because the backend codec already round-trips the mark both ways; verified end to end, typed in the browser and committed as `[the PR](https://…)` in git. `markInputRule` can't express it (it keeps `match[match.length - 1]` as the surviving text, and here the text to keep is the first group), so the replacement is explicit with `removeStoredMark` so following characters don't join the link. Titled links are deliberately left unmatched — Tiptap's `link` mark has no `title` attribute, so converting one would silently drop the title, whereas literal text round-trips intact.

**Current-block highlight removed.** The block under the cursor was tinted with a full-width rounded band, and on an empty block — which you get on every Enter — it wrapped the placeholder in a grey slab that read as a disabled input. Not retuned: the rule's own comment recorded that the subtle tint was imperceptible, so there was no setting between invisible and too loud. This reverses a deliberate choice in #542 (which had a follow-up commit specifically making it *more* visible), on the owner's call.

## Known limitations, carried rather than introduced

**Anything typed while disconnected is dropped**, not merged: the hook builds a fresh doc/awareness pair on every reconnect. Keeping the doc across reconnects would fix it and is only now possible — the server never changes document identity — but that's a behavioural change, not a port.

**Comment anchors use plain-text offsets, not markdown offsets** (`textOffsets.ts`), so highlights, caret activation and deep links drift on formatted content. The span *persisted* when a comment is created is corrected server-side by locating the quoted text in the real markdown — but that correction has a gap now written down in the file: the quote we send is plain text, so a selection containing inline formatting ("hello world" over a source `hello **world**`) matches nothing and falls back to the approximate span. Closing it needs either a markdown serializer client-side or delimiter-insensitive matching in `comment_anchor.py`.

**Links, tables, images and footnotes remain otherwise unauthorable** — the codec reads and writes all of them, but only links now have an input rule.

## Checks

`bun run typecheck` and prettier clean. Built and deployed against the #556 backend on a local stack; co-editing exercised in the browser (typed → synced → checkpointed → committed to git), and each fix above verified there rather than assumed.

`FileView` hardcodes the "pick a template above" placeholder while `WikiPage` picks between that and "Start typing your new document…" based on whether templates exist — so on an install with no templates the copy promises a picker that isn't there. Not fixed: `FileView` doesn't currently know the template list, and threading it through is a call worth making deliberately.


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #558, opened 2026-07-30
feat(ingest): derive the entity-type taxonomy from the wiki

Bo-Onyx:
## Problem

An entity type is a category of thing the wiki tracks — organization, person, software product. Anything that keys facts by entity needs a taxonomy, and the obvious way to get one is for someone to write it down.

That does not survive contact with customers. Nobody arriving with a wiki is going to author a type list first, and a list written for one deployment is wrong for the next.

## Approach

Derive it from the pages. Six stages, reading only the wiki — never an incoming document, because a taxonomy has to describe the corpus it will be applied to:

| | stage | | LLM |
|---|---|---|---|
| 1 | extract | per page, whole page, **no type menu** — supplying one would presuppose the answer | ✅ per page |
| 2 | fold | merge spellings of one thing (case, punctuation, legal suffix, version) | |
| 3 | drop | corpus artifacts — page titles and code symbols are what the wiki is *made of*, not what it tracks | |
| 4 | group | cluster referents by kind | |
| 5 | name | one call per group: name the kind from its observed instances | ✅ per group |
| 6 | merge | one call over the whole taxonomy; per-group naming cannot generalise | ✅ once |

`run_derivation()` is the entry point. `derive()` stays pure, so the pipeline is testable without touching a wiki.

It runs on `automanage_offline_queue`, matching that queue's stated contract: batch, unattended, nobody waiting. **Deliberately not scheduled** — stages 5 and 6 are LLM calls, so re-running can rename a type and orphan anything keyed to the old name.

## Nothing is filtered for being common or rare

Three exclusions were tried and all three removed. Each answered a different question than this pipeline asks:

| removed | it asked | pipeline asks |
|---|---|---|
| minimum sightings | is this **entity** real? | what **kinds** exist? |
| ambient / document-frequency cut | is this referent **discriminating**? | " |
| minimum members per type | does this category have support? | *(right question — the merge already answers it)* |

The sightings filter dropped **66% of referents** on the validation corpus, discarding real companies, products and standards alongside genuine noise. The ambient cut removed 64 on a small sample including several widely-used tools — a tool named on every page is still a tool. The type floor was a no-op whenever the merge worked, while replacing true statements with a bucket: a two-member type says something, `other` says nothing.

Keeping types few and general is the merge step's job, and it does it — **75 types to 9** on the validation corpus. That leaves two parameters where there were six.

## Two decisions worth knowing before changing them

**Folding is lexical, not semantic.** *Lexical* means the comparison looks only at characters — case, punctuation, whitespace, legal suffixes, version markers. Every variant we need to merge differs only in form while denoting the same thing, and none of it requires knowing what the thing is.

Embeddings compare *meaning*, which is the wrong axis: two competing products in one category are semantically close while denoting different things, and a product and its own sub-brand can sit further apart than that pair. So the "same thing, different spelling" and "different thing, similar meaning" bands overlap, and no similarity cut separates them. Containment is avoided for a related reason — a vendor's name usually prefixes its products' names.

**`GROUP_SIMILARITY` is algorithm-specific.** It sets how many groups the naming step sees, and both extremes fail: too low gives a few enormous groups that naming cannot characterise, too high gives singletons it over-specifies. The current value was measured for this leader clustering, which compares against a running centroid that drifts toward a generic average as a group grows — so it tolerates far less permissiveness than a linkage-based method at the "same" number. Retune if the algorithm changes.

## Validation

Against a 137-page corpus: 4,569 mentions → 1,829 referents → **9 types**.

The deterministic stages were then re-run through *this* implementation over the same referents and real embeddings, which is how the clustering mistuning was caught — at the original threshold a single naming call would have been handed 695 referents (38% of the corpus) and asked what kind they share.

## Caveats

- **No consumer in-tree.** This lands the producer; the typed-entity layer that would use it lives in `agent-wiki-evals`. `store_taxonomy()` is a logging placeholder — the eventual home is a *versioned* table, since a re-derivation must not orphan rows keyed under old type names. Read and write land in a follow-up.
- **No home entities are produced.** The organization a wiki is written *by* carries no discriminative signal and the extraction prompt wants to know it. Document frequency was the wrong instrument — it cannot distinguish "whose wiki is this" from "what gets mentioned a lot", and on a small corpus the two are identical. That wants asking each page directly and taking a vote. Until then `load_taxonomy` returns none rather than a guess; the reader already looks for `stats.home_entities` so a producer can fill it without changing that side.
- **The LLM stages have not been run through this implementation** — `llm_settings` reads the database, so a local end-to-end run is not possible. The prompts and parse paths are ported unchanged and the boundary contracts are unit-tested, but a first real run should be watched.

## Review

Five findings from the automated passes, all addressed:

- **naming partition not enforced** — `name_group` accepted whatever it got, so an omitted index silently dropped a referent and a repeated one inflated the support a type is judged on. Now rejected and logged, falling through to the single-type fallback. `_merge_once` already did this; naming was the asymmetry.
- **fingerprint ignored page contents** — hashed path and *length*, so a same-length edit was invisible. That field exists to tell you whether re-deriving is warranted, and it could not. Hashes content now.
- **clusters beyond a 200 cap were dropped** — the cap is gone; naming cost is proportional to a corpus size known before the run.
- **records used `@dataclass`** — `CLAUDE.md` requires `pydantic.BaseModel` for new structured classes and rules out `dataclasses` anywhere. Converted, with `model_dump()` replacing a hand-written serialiser.
- **comments cited one corpus** — parameter comments explained themselves with page counts, referent counts and specific company names from a single validation run. Replaced with the mechanism and the failure modes at either extreme, so the guidance survives a different deployment.


Bo-Onyx:
@greptile

PR #561, opened 2026-07-31
feat(db): store derived entity-type taxonomies, append-only

Bo-Onyx:
Follow-up to #558, which landed the derivation with `store_taxonomy` / `load_taxonomy` as placeholders. This makes them a real read/write seam, so a derivation produces something durable instead of a log line.

## Append-only, with one active row

Entity types are **keys**. Anything recording facts per entity records them under a type name, so a re-derivation that renames a type would orphan every row pointing at the old one. Overwriting a single current value makes that failure silent and unrecoverable.

So rows accumulate and one is flagged `active`:

```
v1  organization, software_product_or_service, person   (superseded, still readable)
v2  organization, software_product, person              (active)
```

`record()` inserts and promotes rather than updating. A consumer that keys by type stores the `version` it used and resolves through `get(version)`, rather than assuming the active taxonomy still means what it meant.

Two things enforce that rather than relying on discipline:

- **a partial unique index** on `active` — the database rejects two rows claiming to be in force, instead of every writer having to remember to clear the previous one
- **deactivate and insert share a transaction** — a half-applied write then either leaves no taxonomy active or fails outright, and the second is much easier to notice

## Schema

```python
entity_taxonomies
  id, version (unique), active, corpus_fingerprint
  types       JSONB   [{name, definition, examples, n_referents, n_docs}, ...]
  provenance  JSONB   models, thresholds
  stats       JSONB   the funnel counts
  triggered_by -> users.id ON DELETE SET NULL
  created_at
```

`types` / `provenance` / `stats` are free-form JSONB deliberately. The shape belongs to the producer (`app.ingest.entity_types`), it is read whole and never queried by field, and pinning it into columns would mean a migration whenever a derivation stage learns to record something new.

`stats` is kept so a human can judge a derivation without re-running it. A merge that collapsed 75 candidate types to 9 looks nothing like one that left 75 standing — and that difference is invisible in the type list alone. It was a real failure mode during development: the merge step silently no-opped, and only the counts revealed it.

## Reading

`load_taxonomy()` reads the active row; `load_taxonomy(version=n)` resolves a specific one.

It falls back to `DEFAULT_TYPES` when no row exists — the same degradation as the relevance scorer without its model file, so a deployment that has never derived keeps working with generic types rather than tailored ones. It now also falls back when a row's entries are unusable, so a malformed taxonomy cannot leave the extractor with an empty type menu.

## Entities are deliberately not stored

Only the taxonomy. Mentions and referents are intermediate — unresolved surface strings with no canonical identity, which a re-derivation invalidates wholesale. Storing them would persist scratch work nothing can key on.

They become durable when resolution exists, and that is a smaller problem once types do: resolution runs *within* a type rather than across all pairs.

## Caveats

- **The tests have not run.** They need Postgres, and neither a local instance nor docker was available. Verified without a database: the DDL SQLAlchemy emits, the migration chain (this is the only revision claiming `c8a4e7d2f6b1` as parent, so no fork), and that both modules import. CI is the first real execution of the 16 tests.
- **No consumer yet.** The typed-entity layer that would key facts by these types still lives in `agent-wiki-evals`. This closes the write path so the producer is complete; the read path exists for whatever lands next.
- **Home entities are still not produced** — unchanged from #558. `load_taxonomy` returns an empty set, and `stats.home_entities` is the field a future per-page ownership vote would fill without touching the reader.

### DIFF
--- a/backend/app/db/entity_taxonomy.py
+++ b/backend/app/db/entity_taxonomy.py
@@ -7,8 +7,8 @@ makes that failure silent and unrecoverable; keeping a history makes a rename a 
 event with the superseded taxonomy still readable.
 
 So ``record()`` inserts and promotes rather than updating, and the previous row stays. A
-consumer that keys by type should store the ``version`` it used alongside, and resolve
-through ``get(version)`` rather than assuming the active one still means what it meant.
+consumer that keys by type should store the taxonomy ``id`` it used alongside, and resolve
+through ``get(id)`` rather than assuming the active one still means what it meant.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #562, opened 2026-07-31
fix(coedit): support code blocks, headings, breaks and tables inside list items

Bo-Onyx:
## Problem

A page whose markdown nests a code block, heading, thematic break or table inside a bullet or blockquote could not be opened **at all** — the body rendered blank with *"This page uses formatting the live editor doesn't support yet."* Page view is the live editor, so this isn't a degraded edit experience; the content is invisible.

`_build_block_sequence` in `markdown_yjs.py` handled exactly three nested block types (paragraph, nested list, blockquote) and raised `NotImplementedError` on anything else, which `api/coedit.py` turns into that refusal frame. A fenced code block inside a bullet is ordinary markdown on a docs page.

## Root cause of the page that surfaced this

The bad markdown was **our own output**. `_serialize_list` indented a task item's continuation blocks by the width of the full `- [x] ` marker (6 columns), but `gfm_parser()` runs no task-list plugin: `[x] ` is the item's own paragraph text, so CommonMark puts the item's content column right after the bullet, at 2. Six columns is four past that — an indented code block on the next parse.

So a task item holding a nested list or a second paragraph was silently rewritten into code the first time anything on that page was checkpointed, and the page then wouldn't open. Round trip, on a real page:

```
- [ ] top            ->  - [ ] top            ->  - [ ] top
  - [x] nested                                        ```
                             - [x] nested             - [x] nested
                                                      ```
```

Continuation now indents to the bullet, so this round-trips.

## Changes

- `_build_block_sequence` gains branches for `fence` / `code_block`, `heading_open`, `hr` and `table_open`, so the nested set matches what `serialize_block` can emit rather than being a subset of it.
- `_serialize_block_sequence` now delegates each child to `serialize_block` instead of keeping its own three-way dispatch — the two directions can't drift into a construct that seeds but won't serialize.
- `_serialize_list` indents a task item's continuation to `- `, not to the checkbox.
- Nested blocks are built from tokens, never from a source slice: token line numbers address the undecorated source, so a slice taken inside a blockquote would carry its `> ` prefixes into the content. Two constructs lose verbatim fidelity as a result — a nested thematic break normalizes to `---`, and a nested table's cells are re-emitted with single-space padding (alignment preserved). Both are now written down in `docs/AGENT_WIKI_MARKDOWN_STANDARD.md` §6.

## Testing

- Seven new codec tests: nested fenced/indented code, nested heading + break, nested table with alignment and with an escaped pipe, blockquote cell re-padding, and the task-item indent regression. Every case asserts a stable (mostly byte-identical) round trip.
- `pytest` green across the backend suite, including `test_coedit_checkpoint`, `test_coedit_rebase` and `test_markdown_splice`.
- Verified directly against the real body of the production page that reported the error: it previously raised `NotImplementedError`, now seeds and round-trips idempotently.

## Follow-up (not in this PR)

The page that surfaced this still holds the mangled 6-space indentation, so it now opens showing its goal tree as a code block — faithful to what the markdown literally says. It needs a one-off dedent back to a real nested list.
<img width="1187" height="825" alt="Screenshot 2026-07-31 at 11 47 48 AM" src="https://github.com/user-attachments/assets/4a1cdf36-b0d6-4658-a962-fd40d11619b5" />

### DIFF
--- a/backend/app/wiki/markdown_yjs.py
+++ b/backend/app/wiki/markdown_yjs.py
@@ -27,13 +27,20 @@ padding, and still achieves the byte-stability goal for every row that isn't
 touched — cells aren't decomposed or individually editable. Thematic break
 and html block stay opaque verbatim, tagged ``_raw="1"``.
 
+A list item or blockquote holds a sequence of these same blocks — not just
+paragraphs and nested lists, but a code block, heading, thematic break or
+table too, since CommonMark nests all of them and a code block inside a
+bullet is ordinary on a docs page. The nested forms are built from tokens
+rather than source slices, so two of them normalize (a thematic break to
+``---``, a table's cell padding to one space) — see ``_build_block_sequence``.
+
 Unrecognized inline constructs (anything this module doesn't have an
 explicit encoder for) raise ``NotImplementedError`` rather than
 silently drop or mis-serialize content — the byte-stability
 requirement this whole engine exists for is only meaningful if failures are
-loud, never silent. Same for a list item that isn't a clean sequence of
-paragraph/list/blockquote children (e.g. a list item containing a table) —
-unsupported, raises rather than mis-encodes.
+loud, never silent. Same for any block construct nested in a list item or
+blockquote that has no branch in ``_build_block_sequence`` — unsupported,
+raises rather than mis-encodes.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


raunakab (approved):

Bo-Onyx:
@greptile

PR #563, opened 2026-07-31
feat(ingest): extract per-page information needs

Bo-Onyx:
Infers what each wiki page **keeps track of** — a stable spec (`description`, `detail_level`) plus a snapshot of the state it currently holds (`current_content`) — rather than extracting the facts on it. A page tracking "current deal status and blockers" has a need whose *shape* is stable even as its content churns, so an incoming document can be judged against what the page maintains instead of diffed against its prose.

First real consumer of the taxonomy from #561.

## What a page is asked to name

A need is **pre-consolidation**, so `need_name` labels the facet the page tracks — "training data schema" — never a topic, and not yet an *aspect* either (an aspect is what the consolidation naming step makes of a need once it can compare across pages). Seven needs naming facets of one model belong to **one** topic, and no single page can see that. So this step names only what it can see, and the stored key says so. Field name aligned with the eval repo's `need_*` schema (#41 there, same reasoning).

## Entities and focus

Each need carries the entities it is **about**, typed from the derived taxonomy rather than a hardcoded list — that is the point of deriving one. Plus a `focus`:

- `SPECIFIC` — the page is about these particular entities and only these (a single customer's account page). Closed set.
- `GENERIC` — the page tracks a *class* and its entities are current instances (a deal tracker with a row per customer). Open set.

`focus` is per **need**, not per page: one page can hold both. An unusable value falls back to `SPECIFIC`, because admitting an entity a page never asked for is worse than omitting one.

`need_kind` and `focus` are `(str, Enum)` and typed on the model, so pydantic refuses an invalid kind at construction rather than only in the parser. This is not hypothetical: measured over 685 needs, **nano put `information_purpose` (14) and `concept` (11) into `need_kind` — 11% of its needs** — which is the eval repo's *topic_kind* axis leaking into the nature axis. Open vocabulary would store those as if they were kinds.

The prompt asks for **every** entity a need is about, with no count hint. An earlier "usually 0-3" suppressed entities on exactly the pages that legitimately name many — a tracker with a row per customer is about every customer in it, and those are the pages the entity-keyed status design depends on.

**No `cadence` field.** It was measured and dropped: populated on only 9–18% of needs; missing from 30/43 timelines under a weak model; leaking onto kinds where it means nothing (16 non-timeline needs on nano, more than its correct uses); and never actually a cadence — 0 of 37 values were a bare rhythm, all prose, and 22/37 restated `detail_level`. Nothing consumed it, and the one use it would serve (staleness) needs a bounded interval, not free text.

## Storage: keyed by doc id, not path

`page_needs` keys on `wiki_doc_ids.id`. Extraction costs an LLM call per page and a move re-keys the doc-id row in place, so **a rename keeps its needs**. Path-keyed, a rename looked like two events — a new page to buy and a vanished one to prune — charging for a reorganization that changed no content. Deliberately unlike `PageEmbedding`, which is path-keyed; that difference is affordable there and isn't here.

No `path` column: after a move the row is intentionally *not* stale, so a denormalized path would go stale on the first rename and never be refreshed. The live path is joined from `wiki_doc_ids` — which also means this table needs no wiki lifecycle hook.

**Current-valued, not append-only.** One row per page; a re-run replaces it. There is no history and no "latest" to resolve, because nothing keys facts by a need, so a replacement has nothing to orphan. That is the exact opposite of `entity_type_taxonomies`, which stays **append-only** with an `active` flag — there, type names are keys that stored needs refer to, so a superseded taxonomy has to remain readable. (A singleton updated in place was explored and reverted: with no superseded row there is nothing to resolve an old label against, so safety would rest entirely on a version counter forcing a full re-extraction.)

It records `entity_type_taxonomy_id`, because entity-type values are only meaningful relative to the taxonomy that defined them; `ON DELETE SET NULL`, so losing a taxonomy costs the ability to resolve type names, not the needs.

Reads return `StoredNeeds` records rather than ORM rows, per CLAUDE.md's *"repos return dicts, not ORM objects"* — same boundary shape as `page_embeddings.PageVector`.

## Also renamed: entity_taxonomies → entity_type_taxonomies

The table holds a taxonomy **of entity types**, but `entity_taxonomies` reads as a taxonomy of entities — an entity registry, which this codebase doesn't have and which is easy to assume it does. Renamed now because the window closes for good: the table landed in #561, has **zero rows in every environment** (verified in prod), and has one consumer, which is unmerged. Once a derivation records a taxonomy, type *names* become keys that stored needs refer to.

## One migration

`f2c9a41e7b06` does both: renames if the old name is present, then creates `page_needs`. These were two revisions and are now one, because they are one change to an operator and the ordering between them was already load-bearing — `page_needs`' FK must be built *after* the rename or it points at a name about to disappear, so splitting them only created a sequence that must not be reordered. Both halves stay guarded, so a database built fresh from the current models (already renamed, already carrying `page_needs`) walks through untouched.

`e1b7c3a95d24` is untouched and still the parent; nothing in this PR has been applied anywhere (prod is at `e1b7c3a95d24`), which is what makes collapsing them safe.

## Incremental, because each page is an LLM call

`content_sha256` / `model` / `entity_type_taxonomy_id` together are the re-extract guard. All three change the output — and a page skipped after a re-derivation would keep entity types the current taxonomy no longer defines, stale in a way nothing downstream could detect. A re-run over an unchanged wiki costs nothing; after one edit, one call; after a rename, nothing. `force=True` re-extracts everything, which is what a prompt change requires, since the prompt is not part of the guard.

Unlike derivation, this is safe to re-run, so it is schedulable in a way derivation is not. It just isn't scheduled yet.

## Three things worth a second look

**The output cap.** Pages are sent whole and unchunked — a page seen in slices cannot be partitioned into the things it tracks, which is exactly what this asks for. `current_content` enumerates a page's entries rather than summarizing them, so responses scale with the page. Measured over a 137-page production wiki, the worst page produced **~8.8k output tokens** (~12.2k under a weaker model) against a p90 of ~2.4k — the client's 4096 default would truncate precisely the dense, many-entity pages this step exists to capture, and truncation arrives as invalid JSON, so the page would retry, truncate again, and be recorded as tracking nothing. Cap is 16384, and truncation is now logged *as* truncation, since its fix is a bigger cap rather than another prompt.

**A prefixed run used to prune globally.** A scoped run passes only its subtree as the live set, so an unscoped prune read every page outside the prefix as deleted and dropped needs that cost a call each — silently, since nothing downstream distinguishes "never extracted" from "wrongly pruned". `prune()` now takes the prefix and bounds itself to it, on a path boundary so scoping to `team` can't sweep `teamwork.md`. Both cases are pinned by tests that fail without the fix.

**`app.tasks.entity_types` was never registered with the worker** (missed in #558). A handler exists only because its module was imported, and the worker's response to a message it cannot handle is to log and **discard** it — so an enqueued derivation would have been swallowed while the caller saw a successful enqueue. Confirmed against the deployed image: the module is absent from its `_TASK_MODULES`. Both that and the new needs task are now registered, and `tests/test_task_registration.py` pins it for every task module by AST-scanning for `.task()` decorators, since neither end of that failure is visible.

## Deletes

`load_all()` excludes tombstoned pages, so a trashed page's needs stop being visible at once rather than at the next extraction — in that window they would still cluster and be reconciled against, routing a fact to a page nobody can see. The row outlives the delete, so trash-then-restore keeps needs already paid for.

## Tests

87 tests. `test_needs.py` covers the parsing contract (closed `need_kind`, off-menu types blanked not dropped, second `primary` demoted, `focus` fallback, atomic per-page returns, the output cap, no entity-count cap). `test_needs_extraction_run.py` runs `run_extraction` end to end against a real wiki repo and database with a stubbed LLM, asserting **call counts** — the unit of cost. `test_page_needs_store.py` covers the guard, the doc-id key (rename, re-mint, prune scope), lifecycle, and the repository boundary. `test_rename_entity_type_taxonomies_migration.py` exercises the rename path a deployed database takes, which fresh test schemas never would.

Full backend suite green; `ruff check` and `basedpyright` clean.


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #565, opened 2026-07-31
fix(automanage): don't count unmanaged files in a sweep's scanned stat

Bo-Onyx:
## Problem

The Auto Organize table reported **193 pages scanned** on a wiki with **148 pages**. Against prod's own path list:

| | count |
|---|---|
| `.md` pages | 148 |
| `.gitkeep` folder markers | 37 |
| `.trigger_*.yaml` | 8 |
| **total files in scope** | **193** |

The extra 45 are objects nobody manages: they never become proposal candidates, and the tree doesn't show them. Counting them reported work on things a user can neither see nor act on.

## They are never candidates — verified per detector

- `body_dup`, `case_collision`, `stub_page`, `template_echo`, `stale_page`, `misplaced_page` — all filter `p.endswith(".md")` before doing anything.
- `empty_folder` — emits `source_paths=[folder]`, the folder path, never its marker.
- `folder_chain` — `_plan_moves` returns `None` on any non-`.md` file under the tail, so a chain holding a trigger or attachment is skipped entirely; its instruction also says to leave `.gitkeep` markers in place.

## Why the scope still keeps them

They're read as facts *about* a folder, not as candidates. Git can't track an empty directory, so a `.gitkeep` **is** the folder: `_maximal_empty_folders` derives folders from `_ancestor_dirs` over the paths it's given, and a folder whose only tracked file is its marker contributes nothing if markers are filtered out — `empty_folder` and `folder_chain` would simply stop finding empty folders. Keeping them costs nothing either; they're already in the single `git ls-files` the sweep runs anyway.

So `run_sweep` still passes every tracked path. Only the reported number changes.

## Change

`paths_scanned` records the count of wiki paths — `.md` today. The column keeps its name because folders join this count when they become scanned units, at which point "paths" still reads true and no stored value changes meaning. The scope size moves into the run log (`scanned 148 pages (193 files in scope)`), where it's the useful cost signal. The admin table's cell label goes from "pages" to "paths" to match.

## Note for the folder follow-up

Folders are not the `.gitkeep` count. Prod has **59** distinct folders but only **37** markers — a folder containing pages needs none. Counting folders as units means deriving them from path prefixes, not counting marker files.

## Historical rows

The two existing rows recorded 193 and will keep reading 193; the file mix behind an old number isn't recoverable from the row, so they aren't back-computed.

## Testing

- New `test_scanned_count_excludes_folder_markers_and_trigger_files`: 5 files in scope (2 pages, 2 markers, 1 trigger), asserts the scope is still 5 and the recorded count is 2.
- Tightened the existing sweep assertion from `>= 3` to exactly `1` — that fixture's three tracked files include two markers.
- Full backend suite green (2182 passed, 2 skipped); ruff, basedpyright, prettier, tsc all clean.

### DIFF
--- a/backend/app/wiki/automanage/runner.py
+++ b/backend/app/wiki/automanage/runner.py
@@ -23,6 +23,7 @@ pages with different audiences are never named together in one proposal.
 Single-path detectors (empty-folder) see the whole scope. Every emitted
 proposal is stamped with the combined audience fingerprint of its path-set.
 """
+
 from __future__ import annotations
 
 import logging
@@ -80,9 +81,7 @@ def _resolve_management(drafts: list[ProposalDraft]) -> dict[str, bool | None]:
     return update_policy.resolve_ai_management_for_paths(paths)
 
 
-def _base_shas(
-    source_paths: list[str], target_paths: list[str]
-) -> dict[str, str] | None:
+def _base_shas(source_paths: list[str], target_paths: list[str]) -> dict[str, str] | None:
     """Drift anchors: every affected path that has history gets one. Paths
     without history are reserved/new names (a rename option, a move
     destination) — nothing to anchor. A draft none of whose paths can be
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #567, opened 2026-07-31
fix(editor): task lists — nested items rendered as columns, checkbox markers escaped away

Bo-Onyx:
Two independent bugs that compound on the same content: a page mixing checkbox items with plain ones renders as a grid of side-by-side columns, and the markers themselves decay into literal text. Reported against a TODO page whose sections had turned into two-column blocks with literal `[x]` text down the middle.

## Layout

`.editor-prose ul[data-type="taskList"] li` is a **descendant** selector, so every nested plain `li` under a task list gets `display: flex` too. A task item wants that — the checkbox sits beside its content — but a plain item renders as `<li><p>text</p><ul>…</ul></li>`, and flexed, its paragraph and its nested list become two columns side by side instead of a list underneath a line of text. On a page that mixes the two, that is most of the page.

Scoped to `> li`. Nested task lists then need an indent of their own: unlike a bullet list they carry no marker to sit under, so they would start at the parent's text column and the nesting would read flat. The added rule matches what the base `ul` rule already gives a nested bullet list.

Page view goes through the same single `.editor-prose` wrapper (`lib/editor/components.tsx`), so this is one change for both surfaces.

## Markers

`_escape_inline_text` escapes every `[` and `]`, so a `[x] ` opening a plain list item comes back out as `\[x\] `. That item's list is one the parse declined to promote — a taskList's children must be uniformly taskItems, so a list mixing marked and unmarked items stays a plain `bulletList` with the marker as ordinary text (`_build_list`), and an ordered list is never promoted at all. The escape then rewrites a marker that GFM readers, and a later uniform version of that same list, still act on into text nothing will ever read as a checkbox again.

It is invisible while it happens — the item renders identically either way — so the first checkpoint of any page with a mixed list quietly spends its checkboxes.

Serialize is now the inverse of parse for that position: a leading marker is emitted bare, and `_ESCAPED_TASK_MARKER_RE` sits next to `_TASK_MARKER_RE` so the pair reads as one rule. Positional and single-shot — brackets anywhere else in the item, including a run that would re-parse as a link, still escape.

## Scope

This does **not** make a mixed list *render* checkboxes; those markers stay literal text until a taskList can hold non-task items, which is a schema change across both sides. It stops the loss, and it leaves the markers in the file for that change to pick up. It also doesn't repair pages already carrying `\[x\]` — that wants a restore from a pre-damage revision.

## Testing

Three new codec tests. Mixed and ordered lists now round-trip **byte-identically** (`once == body`) rather than merely idempotently, which is the property that was broken; the third asserts the exemption doesn't over-reach, and passes with or without the fix by design. Verified the two marker tests fail with the change stubbed out.

`test_markdown_yjs` / `test_markdown_splice` / `test_markdown_blocks` / `test_coedit_checkpoint` pass, ruff + basedpyright + prettier clean.

### DIFF
--- a/backend/app/wiki/markdown_yjs.py
+++ b/backend/app/wiki/markdown_yjs.py
@@ -1018,6 +1018,16 @@ def _serialize_list(node: XmlElement) -> str:
             # (the item renders identically either way). Bare, it re-parses to
             # exactly the text it was serialized from, so the round trip stays
             # byte-stable *and* the marker survives.
+            #
+            # This also normalizes a marker the source deliberately escaped,
+            # and can't do otherwise: markdown-it resolves "\[x\]" to the text
+            # "[x]" before this codec sees a token, so by here the two
+            # spellings are one string and one of them has to be picked for
+            # both. Live is the same choice `_build_list` makes on the other
+            # side of the fork — a list whose items *all* carry an escaped
+            # marker promotes to real taskItems — so the two paths agree
+            # rather than making escaping mean opposite things in a mixed
+            # list and a uniform one.
             body = _ESCAPED_TASK_MARKER_RE.sub(r"[\1]\2", body, count=1)
         lines.append(marker + body)
     return "\n\n".join(lines) + "\n"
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #568, opened 2026-07-31
fix(llm): surface truncated responses instead of returning partial text

Bo-Onyx:
Entity-type derivation lost **every large wiki page** in production. The cause was three layers deep, and the outermost symptom pointed at the wrong layer.

## What actually happened

1. A page's referent list overflowed the client's `DEFAULT_MAX_TOKENS = 4096` (the derivation passed no cap, so it got the default — prod logs show `max_tokens=4096`).
2. OpenAI stopped generating and ended the run with **`response.incomplete`**.
3. `app/llm/providers/openai.py` handled exactly one terminal event — `response.completed` — so that event was **dropped**. The generator ended yielding nothing terminal.
4. `client.complete()` only assigns `stop_reason`/`usage` inside its `done` branch, so it returned the partial text with `stop_reason=''`, zero usage, and **no exception**.
5. `_complete_json` brace-sliced valid-JSON-cut-off-mid-list into invalid JSON and logged `unparseable JSON (N chars)` — naming neither the page nor the reason.

Result: **8 of the 8 largest pages** in a 147-page wiki contributed **zero** referents each, and the log said the model had emitted bad JSON. It hadn't — the model emitted valid JSON and was cut off mid-sentence.

## Confirmed, not inferred

I first concluded "not truncation" from the derivation log showing `status=completed` on every line. That was wrong, and instructively so: the log line lives *inside* the `response.completed` branch, so truncated calls emit **no line at all**. The 19 lines were the successes. Absence of evidence read as evidence of absence.

Settled by wrapping the SDK stream and logging raw event types against the live API:

```
raw SDK event types (non-delta):
   response.incomplete    1        <-- the final event
   ...
delta events: 2021
result: chars=9838 stop_reason='' usage_in=0 usage_out=0
```

The API reported truncation; the caller saw nothing.

## The fix

| file | change |
|---|---|
| `providers/openai.py` | treat `response.incomplete` and `response.failed` as terminal too. **Incomplete** returns with its status and usage — it is a real, usable answer that stopped early. **Failed** raises `LLMError("provider", ...)` carrying the API's error message, like every other provider failure here; returning its partial text even labelled would let a caller persist a broken turn as a finished one, since nothing downstream inspects `stop_reason`. Non-completed statuses log with the API's own `incomplete_details.reason` (`max_output_tokens` / `content_filter`) — the fix differs per reason. |
| `client.py` | a stream ending with **no** terminal event now reports `stop_reason="incomplete"` and warns, instead of presenting partial text as finished. Covers dropped connections and any future status this layer doesn't translate. Uses `"incomplete"` rather than a new sentinel so callers already testing for truncation catch it. |
| `ingest/entity_types.py` | cap raised to 16384 (extraction lists every referent, so output scales with the page); every failure log names the page or group; retry once on **malformed** JSON feeding the error back, but **never** on truncation. |

**Why truncation is never retried:** it's deterministic. Re-asking pays for the same overflowing response twice. What it needs is a bigger cap, not another prompt.

**The cap is a ceiling, not a guarantee.** A large enough page still overflows 16384. That's precisely why the other two changes matter — when it happens it now says so, naming the page, instead of silently yielding nothing. If it recurs, the structural answer is to chunk *mention* extraction: unlike need extraction (where slices can't reveal what a page tracks), "what named things appear in this text?" is genuinely local and the results union cleanly — which is what the fold/dedup stage already does.

## Blast radius

`client.py` and `providers/openai.py` are shared by every LLM call path in the product. Both changes are **additive**: the `completed` path is untouched (pinned by a test), and the client change only fills in a `stop_reason` that was previously empty. Nothing that worked before behaves differently; calls that were silently truncated now report it.

## Tests

13 new. Provider: `incomplete` returns with the right status and usage (and is explicitly *not* treated as a failure), `failed` raises with the provider's detail, and `completed` is unchanged. Client: a stream with no terminal event is reported incomplete and warns. `entity_types`: the cap is passed, truncation is logged as truncation naming the page, truncation doesn't burn the retry, malformed JSON is retried with the reason fed back.

Every fix verified by mutation — reverting it fails a test. Full backend suite green; `ruff check` and `basedpyright` clean.

## After this

Deploy, then a clean derivation run (~147 pages, ~40 min). That's the first run that will tell us what cap the corpus actually needs, since we've never yet seen a complete response for the largest pages.

### DIFF
--- a/backend/app/ingest/entity_types.py
+++ b/backend/app/ingest/entity_types.py
@@ -82,6 +82,17 @@ log = logging.getLogger(__name__)
 GROUP_SIMILARITY = 0.45
 MERGE_ROUNDS = 3  # merge exits on convergence; this only bounds the loop
 
+# Output cap, well above the client's 4096 default. Extraction lists every referent on a page, so
+# the response scales with the page — and on a 205k-char page it overflowed 4096, which cut the
+# JSON off mid-list. Every one of the eight largest pages in a real 147-page wiki failed that way,
+# contributing zero referents each. The cap costs nothing when unused, since billing follows
+# tokens generated. It is a ceiling, not a guarantee: a page can still overflow it, which is why
+# ``_complete_json`` now reports truncation instead of mistaking it for malformed JSON.
+MAX_OUTPUT_TOKENS = 16384
+
+# Retries on MALFORMED output only. Truncation is deterministic, so it is never retried.
+_JSON_RETRIES = 1
+
 # Fallback when no derived artifact is present, so a deployment that has never run the
 # derivation still has something usable. Intentionally generic: these are the types that
 # recur across corpora, not a customer's.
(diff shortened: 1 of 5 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #570, opened 2026-07-31
fix(automanage): show the LLM detectors a tree of pages, not every tracked file

Bo-Onyx:
## Problem

Follow-up to #569. Both LLM detectors build their prompt's "Wiki file tree" from the raw scope:

```python
tree = "\n".join(sorted(p for p in scope.paths))   # stale_page.py:129, misplaced_page.py:143
```

So the model reads every `.gitkeep` marker and `.trigger_*.yaml` in the wiki. On dev-wiki that's **45 of 193 lines — roughly a quarter of the tree** — on every LLM pass of every sweep. None of it is content the model can reason about.

`misplaced_page` had a sharper version of the problem: `_folders()` counts *pages* per folder, so a folder that exists only through a marker is never an accepted destination — yet it appeared in the tree the model was choosing from. The guard at `misplaced_page.py:158` dropped the pick with a warning, i.e. after paying for the round trip. The model was shown options it wasn't allowed to pick.

## Change

Filter both trees through `is_page` (the util from #569). Folders holding pages still appear — a page path carries its prefix — and marker-only folders drop out, which is exactly the set `misplaced_page` refuses anyway.

The **scope is unchanged**. `empty_folder` and `folder_chain` still read the unfiltered list, because a marker is their only evidence a folder exists. This is about what the model is shown, not what the sweep covers.

## Testing

- New `test_prompt_tree_holds_pages_only`: seeds a scope holding a marker-only folder and a folder-scoped trigger, captures the prompt through the existing `client.complete` stub, and asserts pages are present while `.gitkeep`, the trigger YAML, and the marker-only folder are not.
- Confirmed the test isn't vacuous — reverted the change and watched it fail on the `.gitkeep` assertion before restoring.
- Full backend suite green (2276 passed, 2 skipped); ruff and basedpyright clean.

## Not verified

The prompt input changes, so the detectors deserve a live run against the real model before this is trusted in production — the same live verification #532 and #543 each got. I don't have provider credentials in this environment, so that check is outstanding.

### DIFF
--- a/backend/app/wiki/automanage/detectors/misplaced_page.py
+++ b/backend/app/wiki/automanage/detectors/misplaced_page.py
@@ -129,7 +129,13 @@ class _MisplacedPageDetector:
     def _agent_pass(
         self, scope: Scope, candidates: list[str]
     ) -> list[dict[str, str]]:
-        tree = "\n".join(sorted(p for p in scope.paths))
+        # Pages only. The scope also carries the `.gitkeep` markers that
+        # materialize folders and any `.trigger_*.yaml` — the folder
+        # detectors need those, but they are not content the model can
+        # reason about, and a marker-only folder is not a destination
+        # `_folders` would accept anyway. Folders holding pages still
+        # appear, since a page path carries its prefix.
+        tree = "\n".join(sorted(p for p in scope.paths if is_page(p)))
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
@greptile

PR #578, opened 2026-08-03
perf(ingest): extract and name concurrently in the entity-type derivation

Bo-Onyx:
The entity-type derivation ran **fully sequentially** — no `ThreadPoolExecutor`, no `concurrent`, no `asyncio` anywhere in `app/ingest/entity_types.py`. 147 blocking LLM calls in series at ~55s each.

In production that meant **~1.5 hours**, and a pod restart killed the run past **105 of 147 pages** before `record()` could write anything — so every one of those calls was discarded (~471k input / ~309k output tokens). The calls are independent, so that wall clock was pure serialization.

A bounded pool of 8 brings it to roughly **15 minutes**, which also shrinks the window an interruption can land in.

## What is parallelised

| stage | unit | count | parallel? |
|---|---|---|---|
| **EXTRACT** | per page | 147 calls | **yes** |
| **NAME** | per candidate group | ~220 calls at `GROUP_SIMILARITY = 0.45` | **yes** |
| **MERGE** | the whole taxonomy, iterated to convergence | 1 per round | no — inherently global |

Naming is not the small stage its name suggests: at the tuned threshold it's *more* calls than extraction, so leaving it sequential would leave most of the wall clock in place.

## The part worth reviewing: order is load-bearing

Concurrency here must not change the output, and a plain as-completed loop **would**:

- `_leader_cluster` assigns each referent to the **first** centroid it matches.
- The naming collapse keeps **first-seen** examples.

So a reordered result set is a *different taxonomy from the same corpus* — silently, with no error and no way to notice. `_map_ordered` uses `Executor.map`, which yields by submission index rather than completion, so a parallel run is reproducible and identical to a sequential one.

The test for this makes completion order the exact **reverse** of input order; collecting as-completed fails it.

## Failures stay per-item

The module's rule is that one bad page must not abort a corpus-wide derivation. A raised exception inside a pool violates that — it propagates out of `map` and takes every other page's paid-for work with it. `_guarded` absorbs it and returns the same empty result the sequential path produced, so a single failure costs one page.

## The worker count is an operator knob

Set by **`ENTITY_TYPE_DERIVE_WORKERS`** (default 8), via `Config` like every other ingest knob. It is not a build-time constant because the governing constraint is the **provider's rate limit**, and this pool runs *inside* a queue task — the queue's own concurrency is itself a `ThreadPoolExecutor`, so effective LLM concurrency is that worker count **multiplied** by this. A deployment on a lower tier has to be able to turn it down without rebuilding the image.

Read per call rather than captured at import, clamped to a floor of 1 (0 would mean a pool of no workers — a derivation reporting success having done nothing), and a non-integer value raises at boot rather than silently falling back to a default nobody chose.

Eight matches the eval harness this module was ported from (`ThreadPoolExecutor(max_workers=8)`, `DEFAULT_WORKERS = 8`). The parallelism was in that harness and did not survive the port — which is also why my first estimate of this job's duration was wrong by an order of magnitude, since it was reasoned from the harness's behaviour rather than the product's.

## On moving fan-out into the queue instead

Raised in review, and worth answering explicitly: it isn't available here. The queue has **no group/chord/barrier primitive**, and the derivation needs *every* page before fold → group → name → merge can run. Queue-level fan-out would therefore mean building a join — a state table, a completion counter, and a "last worker continues" handoff — for a job that is unscheduled and occasional. There is also precedent for bounded in-process pools in app code (`web/firecrawl.py`), and the queue's own consumer concurrency is a thread pool.

The pool here is bounded, short-lived, scoped to a single task, and now operator-governed, which addresses the part of that concern that actually bites.

## Not affected: prompt caching

Worth stating because it's the obvious objection to fanning out. Caching was already at **zero** — every call in both prod runs logged `cached=0`. OpenAI needs a shared prefix of ≥1024 tokens and `entity_types.extract` is ~396; everything after it is the page body, which is unique per call. Parallelism doesn't change eligibility either way.

(Separately, that gap is a real cost lever — input dominates spend and the prefix is ~600 tokens short of cacheable. Out of scope here; worth measuring on the next full run, which will produce real `cached=` figures.)

## Tests

11 new, covering the properties rather than the mechanism: input order preserved despite reversed completion order, concurrency actually occurs, concurrency is bounded, one failing item doesn't abort the rest, progress reports every item, an empty corpus starts no pool. Plus the knob: it defaults to 8, the env var lowers it, a lowered value really bounds the pool, 0 clamps to sequential rather than stalling, and a bad value fails at boot.

Verified by mutation — collecting as-completed, dropping the guard, and forcing `workers=1` each fail a test. Full backend suite green; `ruff check` and basedpyright clean (581 files).

## After this

Deploy, then run the derivation as a **Kubernetes Job** rather than a detached process in the web pod. At ~15 minutes the interruption risk mostly disappears, but the web pod is still the wrong host — that's what lost the last run, along with its log.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -77,6 +77,11 @@ class Config(BaseModel):
     # no-op. Populates the store for the cosine cold-start / two-tower warm
     # filter that lands in later phases.
     ingest_embed_model: str
+    # How many entity-type extraction / naming LLM calls run at once during a derivation.
+    # Operator-tunable because the ceiling is the PROVIDER's rate limit, not this code: each
+    # call can carry ~47k input tokens, and this pool sits inside a queue task, so the real
+    # concurrency is the queue's own worker count multiplied by this. Lower it on a lower tier.
+    entity_type_derive_workers: int = 8
 
     # Relevance filter (drops irrelevant candidate pages before the LLM reconcile).
     # Empty model path => cosine cold-start filter; a path to an exported two-tower
@@ -208,6 +213,7 @@ def load_config() -> Config:
         ingest_bm25_limit=_positive_int("INGEST_BM25_LIMIT", 20),
         ingest_irrelevant_stop_n=_positive_int("INGEST_IRRELEVANT_STOP_N", 2),
         ingest_embed_model=os.environ.get("INGEST_EMBED_MODEL", "text-embedding-3-small"),
+        entity_type_derive_workers=_positive_int("ENTITY_TYPE_DERIVE_WORKERS", 8),
         ingest_relevance_two_tower_model_path=os.environ.get("INGEST_RELEVANCE_TWO_TOWER_MODEL_PATH", ""),
         # Cosine cold-start default: the study's 85%-filter operating point.
         ingest_relevance_cosine_threshold=_nonneg_float(
--- a/backend/app/ingest/entity_types.py
+++ b/backend/app/ingest/entity_types.py
@@ -51,9 +51,10 @@ import logging
 import math
 import re
 from collections import Counter, OrderedDict
+from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timezone
 from collections.abc import Callable
-from typing import Any, cast
+from typing import Any, TypeVar, cast
 
 from pydantic import BaseModel, Field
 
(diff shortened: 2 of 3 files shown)


Bo-Onyx:
@greptile

PR #579, opened 2026-08-03
fix(wiki): one identity colour for a peer's caret and their presence chip

Bo-Onyx:
Confirmed — and the mismatch isn't intermittent, it's structural.

## What was wrong

A peer appears twice on a page: as a caret in the document, and as the ring on their presence chip. Those colours were derived two different ways:

| | palette | keyed by |
|---|---|---|
| caret (`presence.ts`) | `PEER_COLORS` — 8 raw hex (`#e5484d`, `#0090ff`, …) | hash of `userId` |
| chip ring (`PresenceAvatars.tsx:348`) | `USER_COLORS` — 5 Opal neon vars | **index in the participant list** |

Different palettes *and* different keys, so any agreement was a coincidence of similar hue rather than a match — which is why it read as "sometimes different."

**Keying the ring on list position is a second defect.** The list filters out the current user and is in participant order, so a peer's ring colour changed when somebody *else* joined or left ahead of them. Demonstrated against the two implementations:

```
caret vs ring, per user:
  8840e472-user-bo     caret=lime     oldRing=cyan
  c1a2-user-joachim    caret=yellow   oldRing=yellow   ← coincidence
  f9d3-user-nik        caret=magenta  oldRing=lime
  aa11-user-dane       caret=lime     oldRing=magenta

same peer, roster changes (nik):
  old ring: lime -> yellow      ← peer did nothing; someone ahead of them left
  new ring: magenta -> magenta
```

(That comparison holds the palette constant to isolate the keying difference; in the shipped code the two palettes differed as well, so the disagreement was near-total.)

A colour whose only job is "this caret belongs to that chip" cannot move when the roster around it moves.

## Fix

One module, `lib/editor/identityColor.ts`, owns the palette and the keying rule; the caret and the chip both call its `colorFor(userId)`. `presence.ts` re-exports it so existing importers are unaffected.

The surviving palette is the **chip's existing Opal tokens**, not the caret's hex literals — one list to change, and no raw colour values (CLAUDE.md). Keyed on `userId`, never on position.

## Trade-off, and one thing for Nik

The caret palette goes from 8 entries to 5, so distinct users collide slightly sooner. A brief ambiguity between two peers seems clearly better than a colour the reader can't rely on at all — but if you'd rather keep 8, the fix is to add three more Opal tokens to the one list (`--blue-50`, `--orange-50`, `--red-50` all resolve).

Worth a look from Nik since he owns this surface: two of the inherited chip tokens are light (`--neon-yellow-50` `#ece600`, `--neon-cyan-50` `#00ebea`) and now render as caret bars too, where they'll be fainter on a light background than the hex values they replace. Swapping specific hues is a one-line edit now that there's a single list.

`tsc` and prettier clean.

Screenshot 2026-08-03 at 11 32 25 AM

### DIFF
--- a/frontend/src/components/wiki/PresenceAvatars.tsx
+++ b/frontend/src/components/wiki/PresenceAvatars.tsx
@@ -24,6 +24,7 @@ import {
   PolicyPopover,
   type OpenUpdatesPanelOpts,
 } from "@/components/wiki/policyPanels";
+import { colorFor } from "@/lib/editor/identityColor";
 import { relativeTime } from "@/lib/users";
 import { fetchFileHistory } from "@/lib/wiki/svc";
 import { useUpdateHealth } from "@/lib/wiki/hooks";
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #580, opened 2026-08-03
feat(ingest): derive the taxonomy with the ingestion model

Bo-Onyx:
The taxonomy derivation ran on the main model. It's ingest-side work, so it should use the model an admin picks for the ingestion pipeline.

`run_derivation` now defaults to `llm_settings.ingest_selector_model`. Unset means no cheaper model was nominated, so it falls through to the deployment default; an explicit `model=` still wins. `derive_entity_types(model=...)` passes through for one-off runs.

**Why mini is enough here:** the validated 9-type taxonomy for the reference wiki was produced by `gpt-5.4-mini` — same corpus, one model for every stage.

**Why it's also safer:** on the production run with the main model (`gpt-5.5`), one page produced **14,453 output tokens against the 16,384 cap**. Truncation in this stage drops that page's referents outright rather than degrading them, so a less verbose model widens a margin that is currently thin.

## Tests

3 new: the ingestion model is used by default, an explicit model overrides it, and an unset ingestion model passes `None` (not `""`, which would read as configured-but-blank). Verified by mutation — dropping the default fails the first.

`ruff check` and basedpyright clean.

## Note

Need extraction (`app/ingest/needs.py`) still resolves the **main** model. It's also ingest-side, so the same argument may apply — but it's a separate change with its own cost/quality trade-off, and switching it would make every stored page stale and force a full re-extraction. Left alone deliberately.

### DIFF
--- a/backend/app/ingest/entity_types.py
+++ b/backend/app/ingest/entity_types.py
@@ -61,6 +61,7 @@ from pydantic import BaseModel, Field
 from app.config import CONFIG
 from app.db import entity_type_taxonomy
 from app.llm import client, embeddings
+from app.llm.settings import get as get_llm_settings
 from app.llm.prompts import load_prompt
 from app.wiki import filesystem, git as wiki_git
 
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #582, opened 2026-08-03
fix(ingest): accept partial naming responses instead of discarding them

Bo-Onyx:
The naming stage enforced a strict partition and treated any violation as no answer at all. On the production derivation that rejected the **three largest groups — 261, 119 and 94 referents — each for a single uncovered member**:

```
naming covered 260 of 261 member(s); ignoring response
naming covered 118 of 119 member(s); ignoring response
naming covered  93 of  94 member(s); ignoring response
```

Those groups fell through to the whole-group placeholder (`unnamed_261`, definition `"(naming failed)"`), so the merge step had to place **474 referents** from example lists with no definition to constrain them.

**That is visible in the recorded taxonomy.** `software` is the largest type at 1012 referents, and its examples include `OpenAI` and `Braintrust` (organizations) and `Bo` (a person). The vocabulary's biggest type is absorbing referents that belong elsewhere.

The guarding intent was right; the remedy was inverted. Discarding the response drops **all 261** members to avoid silently dropping **1**.

## The fix

Both ways the partition can break now take what is valid:

| violation | before | after |
|---|---|---|
| **omitted** members | discard the whole response | keep the named types; carry the uncovered members as their own small remainder, so nothing is dropped silently |
| **repeated** members | discard the whole response | first claim wins, later duplicates dropped — support counts stay exact, which is what the original guard protected |

Merge can place a handful of leftover referents from examples far better than it can place an entire group.

This is the same conclusion `merge_types` already reached one stage over — its docstring notes *"a single conflict used to discard the entire response."* Naming kept the old rule.

## Placeholder names are per group

Raised in review: `derive` collapses types that **share a name**, on the premise that the model named them identically — a premise a placeholder doesn't carry. Sized names meant every group leaving one unassigned member produced `unnamed_1`, so the collapse pooled unrelated referents and examples into one type and merge then placed them together.

Partial acceptance makes small remainders common, so this would have fired often — a miniature of the very misplacement this PR set out to fix. Names now derive from the first member's canonical, which folding has already made distinct. The whole-group fallback had the same flaw and is fixed too.

## Tests

7 new: one uncovered member no longer discards the group; a full partition is unchanged; a repeated member keeps the first assignment; a genuinely unusable response still falls back to the whole-group placeholder; referent counts sum to the group with nothing double-counted or lost; and remainders from different groups (and whole-group fallbacks) get different names.

Verified by mutation — restoring either all-or-nothing rule, or the sized placeholder names, fails a test. `ruff check` and basedpyright clean.

## After this

Re-derive. It's cheap now (~8 minutes on `gpt-5.6-luna` after #578), and re-deriving later would invalidate every extracted need anyway, since the taxonomy id is part of need extraction's staleness guard. Better to settle the vocabulary before anything labels against it.

Worth comparing the re-derived taxonomy against the current one: if `software` shrinks and `organization`/`person` grow, that confirms this was the cause of the misplacement.

### DIFF
--- a/backend/app/ingest/entity_types.py
+++ b/backend/app/ingest/entity_types.py
@@ -414,6 +414,18 @@ def _member_indices(entry: dict[str, Any], upper: int) -> list[int]:
     return out
 
 
+def _placeholder_name(members: list[Referent]) -> str:
+    """A name unique to THESE members.
+
+    ``derive`` collapses types that share a name, on the premise that the model named them
+    identically. A placeholder does not carry that premise, so two groups each leaving one
+    unassigned member must not become one type with their referents and examples pooled.
+    Canonicals are distinct after folding, so the first member's is enough to keep them apart.
+    """
+    slug = _TOKENS.sub("_", members[0].canonical.lower()).strip("_")
+    return f"unnamed_{slug or 'group'}"[:60]
+
+
 def name_group(group: list[Referent], *, model: str | None = None) -> list[EntityType]:
     """Name the kind a group shares. May split a group that turns out to be mixed."""
     system = load_prompt("entity_types.name")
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #583, opened 2026-08-03
fix(ingest): let the merge stage actually reach convergence

Bo-Onyx:
The recorded taxonomy has 23 types, including `identifier` (**1** referent), `metric` (2) and `technical_concept` (2, `Hyrum's Law`) — types the merge prompt explicitly asks to fold: *"a type with 1 referent almost always belongs inside a larger one"*, *"PREFER FEW, GENERAL types"*.

The prompt wasn't being ignored. The loop was stopping early.

## Cause

`merge_types`' docstring says it *"iterates to convergence"*. The code said:

```python
for _ in range(MERGE_ROUNDS):   # MERGE_ROUNDS = 3
```

Naming emitted **204** candidate types on the production corpus, and the loop stopped mid-descent at 23. The eval's 75 candidates converged inside 3 rounds, which is why the cap never surfaced — it only binds once the input gets large, and `gpt-5.6-luna` emits ~2.7× more candidates than `gpt-5.4-mini`.

Worse, **the two outcomes were indistinguishable from outside.** There was no per-round logging, so "converged at 23" and "still collapsing when the cap hit" looked identical. That's why this needed a taxonomy inspection to spot rather than a log line.

## Fix

- `MERGE_ROUNDS` 3 → **8**, as a safety bound rather than a working limit. The loop still exits the moment a round collapses nothing, and one round is one LLM call, so a generous cap costs nothing when convergence is early.
- **Log every round** (`merge round 2: 40 -> 20 type(s)`), and log convergence with the full trace.
- **Warn when the cap binds**, so a still-collapsing taxonomy announces itself instead of passing as a stable answer.

## A failed round is not convergence

Raised in review, and it undercut the fix above: `_merge_once` returned its input unchanged on a provider error, a truncated response, or an unusable payload — indistinguishable from a round that ran and merged nothing. With the new convergence check, a **failure would have logged "merge converged"** and the partially merged taxonomy would be persisted as a stable answer.

It now returns `None` for a failure. `merge_types` stops with an explicit `NOT converged` warning and keeps the progress earlier rounds made. Fewer than three types remains a clean no-op, not a failure.

## One prompt addition

Judge a type by its **members**, not its label. The recorded taxonomy had deployment regions (`us-gov-west-1`, `us-east-2`, `America/New_York`) under `geographic_entity`, and companies' domains (`badger.blue`, `founda.com`) under `website_or_online_site` — plausible labels sitting on members that belong elsewhere.

Stated as a principle, not as corpus-specific corrections: the point of deriving a taxonomy is that it isn't hand-authored.

Scoped to what the response can actually express, per review: merge reassigns **whole type indices**, not individual referents, so the rule asks for a whole-type judgement and says to leave a genuinely mixed type alone rather than move it somewhere only some members fit. Every example is now listed rather than the first five, so that judgement has the evidence to make it.

## Tests

6 new: merging continues while the count falls; it stops the first round nothing collapses; hitting the cap warns; each round is reported; a failed round reports NOT converged and keeps earlier progress; and too-few-types is not treated as a failure. Verified by mutation — restoring the 3-round cap, demoting the cap warning, or returning the input on failure each fails a test.

`ruff check` and basedpyright clean.

## What this does not fix

Two problems in that taxonomy are **not** the merge loop, and I'd rather name them than imply this is a complete answer:

- **Folding misses** — `Onyx` vs `Onyx Cloud`, `Evan` vs `Evan Lohn`, `agent-wiki-evals` vs `onyx-dot-app/agent-wiki-evals`. Deterministic and model-independent: `normalize_surface` strips legal suffixes and version tokens and nothing else. That's the documented lexical-vs-semantic tradeoff, and these are its cost.
- **Run-to-run instability** — two runs over an identical corpus gave 18 and 23 types with only 6 names shared, and extraction varied 6,120 vs 5,650 mentions. Sampling non-determinism, not a code path. Convergence should reduce its *visible* effect by collapsing harder, but won't remove it.

Re-derive after this deploys and compare: expect fewer types, no single-referent types, and a `merge converged after N round(s)` line confirming the loop finished on its own terms.


Bo-Onyx:
@greptile

PR #585, opened 2026-08-03
fix(coedit): stop the session purge from cutting a page's Yjs lineage

Bo-Onyx:
## The bug

A page duplicated itself repeatedly in production, twice — the second time
with #575 already deployed, so #575's fix was necessary but not sufficient.

A page's document identity **is** its Yjs lineage: Yjs merges by item id
(`client_id`, `clock`) and never by content, and `seed_doc_from_markdown`
builds a bare `Doc()` with a random client id. Two documents independently
seeded from the *same* markdown therefore share nothing, and merging them
yields the content twice.

#575 made a reconnect reuse the closed session so the lineage survives. But
reuse can only find a row that still **exists**, and the every-minute scan
deletes the row in the same pass that closes it:

```
expire_stale_participants -> close_abandoned_sessions -> purge_viewer_sessions
```

`purge_viewer_sessions` targets `ydoc_seq == 0` — never edited — which is
precisely a *viewer's* row, and every page view mints a session. So a lapsed
heartbeat on a still-open tab closed the session, deleted it, and the
client's reconnect ~3s later got a fresh `Doc()`: a new lineage facing the
document the browser was still holding. The client answers SYNC_STEP1 with
all of it, nothing dedupes, the next checkpoint commits the page twice, and
the following session seeds from that — geometric.

## Changes

**`purge_viewer_sessions` gains `retain_seconds` and skips the newest closed
row per path.** The age gate covers the reconnect window; keeping the newest
row covers a client that reconnects much later (a suspended laptop), and is
bounded at one row per page.

**`drop_restated_blocks`, a second checkpoint net** for the shape the
existing duplicate-id check structurally cannot see. Once a checkpoint
restamps a collided doc, its children carry *distinct* sequential ids, so no
id repeats — while `checkpoint_body` re-serializes every child whose id
`base_body` doesn't know (`orig is None`, read as "freshly typed") onto the
end of the page. That produced the observed commit shape: 87 insertions, 0
deletions, one hunk at EOF. The new check keys on content instead — children
restating a base block that is still being emitted — behind both a minimum
count and a minimum fraction, so retyping one line the page already has is
never touched. It deletes rather than refuses, like its sibling, so the
checkpoint's broadcast converges the browsers holding the duplication.

## What this does not do

Neither change removes the underlying fault: a lineage is minted ad hoc by
whichever code path needs a document. One lineage per path — moving the
snapshot off `coedit_sessions` onto a per-path row seeded exactly once — is
the structural fix, written up in the design doc and not scheduled yet.

## Testing

Full backend suite green. Four new tests: the collision shape the id check
misses (a restamped doc with distinct ids), a healthy-doc no-op, the floors
holding off a legitimately repeated line, and purge retention (recent rows
retained, newest-per-path kept regardless of age, older ones still purged).

### DIFF
--- a/backend/app/tasks/coedit_checkpoint.py
+++ b/backend/app/tasks/coedit_checkpoint.py
@@ -45,6 +45,13 @@ _MAX_INTERVAL_SECONDS = 900
 # Four missed 15-second heartbeats distinguish a departed participant from
 # ordinary event-loop lag.
 _PARTICIPANT_STALE_SECONDS = 60
+# How long a closed viewer session stays purgeable-exempt. A reconnecting
+# client adopts its old row's Yjs lineage, and only a row that still exists
+# can be adopted, so this has to outlast a reconnect by a wide margin — the
+# client retries every ~3s, but a suspended laptop resumes far later. Held
+# well above that; the newest closed row per path is kept regardless of age,
+# so this window only bounds how long the *older* ones linger.
+_VIEWER_SESSION_RETAIN_SECONDS = 3600
 
 
 @coedit_queue.task()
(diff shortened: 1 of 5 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #586, opened 2026-08-03
feat(editor): merge active agents into the presence avatar cluster

Bo-Onyx:
Removes the "Active agents" / "No agents active" bar and merges everything it surfaced into the header's presence avatar cluster, per the design direction for the presence section.

## What changed

**Bar removal**
- `ActiveAgentsBar` / `ActiveAgentRow` / `ActiveSessionRow` and the sub-header portal slot (`AgentsBarHost` + its context in `WikiHeaderActionsProvider`) are gone.
- `ActiveSessionsList` (+ module CSS) deleted — it was dead code, mounted nowhere.

**Session-granular presence**
- The avatar roster is now one chip per live connection (`CoeditPeer` gains `client_id`), not one per user. The only session ever hidden is the tab you're looking from: your own other tabs, your launcher sessions, and your agents all render.
- Users with no live connection keep a single per-user viewer chip, as before.

**Identity colors per session**
- Each connection already advertises its caret color via awareness; that key moves from `colorFor(userId)` to `sessionColorFor(userId, clientId)`, and chips now display the peer's *advertised* color (`CoeditPeer.color`) instead of recomputing — so caret↔chip agreement (the invariant `identityColor` exists for) holds while the same user's two tabs get two distinguishable colors. Caret-less chips (viewers, launcher sessions, standalone agents) key on their own stable ids.

**Agents + launcher sessions in the cluster**
- Agent activity badges onto the user's launcher-session chip when one exists (that chip usually *is* the agent), else their first chip, else seats its own chip — which is how your own agent stays visible while your tab's chip is hidden.
- Your launcher sessions render as your avatar badged with the tool glyph. The presence card keeps the actions the old bar had: **Close agent session**, and **Open Craft** for a ready Onyx Craft session. Failed Craft sessions no longer linger in UI — the `CraftNotifier` toast owns failures.

## Notes for review
- Mobile: the old bar rendered on mobile; `PresenceAvatars` is desktop-only, so mobile currently has no agent-presence surface. Deliberate for now — flag if mobile needs a treatment.
- A reconnect re-rolls a session's color (awareness client id changes), on caret and chip together.
- Verified with `tsc` + prettier; presence flows exercised via the existing structure only — a live two-tab smoke test on the deployed stack is worth doing before merge.
Screenshot 2026-08-03 at 2 30 47 PM


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #587, opened 2026-08-03
fix(coedit): bound the wait for a checkpoint ack

Bo-Onyx:
## The bug

A tab sat on "Saving…" for 16 minutes. Investigation showed the save had
worked fine:

- the session was **clean** — `ydoc_seq == ydoc_checkpointed_seq`, last
  checkpoint 20:45:27
- the coedit queue was healthy — `xlen=1`, `pending=0`, consumers active
- **no** errors, warnings, or tracebacks in either pod for the whole hour, and
  no `busy; skipping` line, so not even lock contention

The checkpoint committed normally in the usual few hundred milliseconds. The
`checkpoint_result` frame just never reached the browser.

`checkpointSession`'s promise settles on exactly two events — that frame, or
`ws.onclose` rejecting everything pending. Socket open + ack missing = pending
forever.

And the label was the smaller half of it. `checkpoint()` sets
`saveInFlight.current = true` before awaiting and guards with
`if (!canWrite || saveInFlight.current) return`, so a hung promise means **that
tab never autosaves again** for the life of the connection — silently, with no
error and no retry, because the retry only arms on rejection. Work still
reaches git, but only through the server's 5-min-idle / 15-min-overdue scan.

## Changes

**A 60s deadline on the wait.** Sized off the two legitimately slow paths so it
can only fire on something genuinely wrong, never on a slow-but-working save:
the server's `_CHECKPOINT_LOCK_TIMEOUT_MS` (30s — how long a duplicate
checkpoint waits for the one in progress), plus room for the AI merge that
resolves an overlap with an agent commit, which the backend measures in
seconds. A comment ties the two constants together since they can't be shared
across the stack.

**A separate retry budget for timeouts (1).** `SAVE_RETRY_LIMIT = 10` is sized
for "not connected" during a reconnect, which fails instantly — 10 × 3s is a
sensible 30-second window there. At a minute per attempt the same budget means
~11 minutes of alternating "Saving…"/"Couldn't save". One retry bounds it to
about two minutes and then leaves an honest error.

**An honest message.** A missing ack says nothing about whether the commit
happened, so the indicator no longer claims failure; and once the budget is
spent it says what actually happens next — the periodic scan commits it. Giving
up delays the commit, it doesn't lose it: the edits are already in
`coedit_updates`.

**Control-frame delivery logging.** `_deliver_local` silently drops a frame
when no connection for that session is registered in the process — normal for a
task worker (it publishes; the socket lives in the web process, reached via the
bus), but also exactly how an ack goes missing, and it left no trace at all.

## What this does not do

**It does not explain why the frame vanished.** The server side is sound: the
task acks in a `finally`, so even the busy-skip path publishes, and the queue
wasn't backed up. Whether the frame was never fanned out or arrived without
matching is unknown, and after the fact there's nothing to inspect. The logging
above is what makes the next occurrence answerable; I'd rather add the trace
than guess.

What this *does* fix is the amplification — a single lost frame permanently
breaking a tab, versus a blip that self-heals in seconds.

## Testing

Backend: ruff, basedpyright, 130 coedit tests pass. Frontend: `tsc --noEmit`
and prettier clean.

The timeout path itself is **not** covered by an automated test — the frontend
has no test runner configured (`CLAUDE.md` lists Vitest as "when needed"), and
adding one is outside this change. The behaviour wants a runtime check against
a suppressed ack before this is trusted.

### DIFF
--- a/backend/app/wiki/coedit_channel.py
+++ b/backend/app/wiki/coedit_channel.py
@@ -164,16 +164,24 @@ def _deliver_local(coedit_session_id: int, frame: ControlFrame) -> None:
         q.put_nowait(frame)
         notify()
         delivered += 1
-    # A process with no connection for this session is the normal case for a
-    # task worker (it publishes; the socket lives in the web process, reached
-    # via the bus), so this is not an error on its own. It is logged because a
-    # frame delivered to nobody *anywhere* is otherwise invisible: a
+    # A frame delivered to nobody *anywhere* is otherwise invisible: a
     # checkpoint ack that never lands leaves the client's save promise pending,
-    # and until this line there was no trace to distinguish "never published"
-    # from "published, never delivered".
-    log.info(
+    # and without this there is no trace to distinguish "never published" from
+    # "published, never delivered". A process holding no connection for the
+    # session is the normal case rather than a fault — a task worker publishes
+    # while the socket lives in the web process, reached via the bus — so this
+    # records the counts and judges nothing.
+    #
+    # Only the checkpoint ack is worth INFO: it is the one whose loss is known
+    # to strand a client, and it is rare (one per edit burst). Presence frames
+    # go out on every join and leave, which would be noise at this level and
+    # would bury the signal. They stay at DEBUG, where the whole trace is still
+    # available when someone is looking for it.
+    frame_type = frame.get("type", "?")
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
@greptile

PR #590, opened 2026-08-03
fix(wiki): update-history diff uses the shared edge scrollbar

Bo-Onyx:
The update-history diff pane scrolled with its native scrollbar — a full-height rail plus thumb — while the live editor shows a thumb-only indicator. Same document, two scrollbar treatments.

## What changed

- `EditorEdgeScrollbar` is generalized into `EdgeScrollbar` (one module for every doc surface): it now drives any `EdgeScrollTarget` — an interface `CoeditorHandle` already satisfies structurally — and `useElementScrollTarget` adapts a plain DOM scroller to it (scroll listener + ResizeObserver on the container and its content wrapper, re-aimed when the content is swapped for a different commit's diff).
- `DiffView` hides its native scrollbar and mounts the shared thumb at its right edge. The `scrollbar-gutter: stable` reservation goes with the native bar — with no bar there's no width shift between overflowing and non-overflowing diffs.
- Thumb color unified to `--text-02` (hover `--text-04`) on both surfaces — the darker tone the native bar had, now consistent everywhere.

Verified in the browser: history mode shows a thumb with no rail, same geometry and color as the editor's docked thumb; track click/drag work through the same pointer-capture path as before.


Bo-Onyx:
@greptile

Bo-Onyx:
Follow-ups pushed on review: thumb lightened to Google-Docs weight (`--alpha-grey-100-30`, hover `--text-02`), and the folder view (`DirView`) + landing page (`WikiHome`) now share the same EdgeScrollbar with their native bars hidden — every wiki scroll surface shows one consistent thumb, no rail.

Bo-Onyx:
@greptile

PR #591, opened 2026-08-03
perf(ingest): extract needs concurrently, on the ingestion model

Bo-Onyx:
Readies need extraction for its first production run. Two changes.

## Concurrency

Extraction was sequential — 148 pages at ~45s each is roughly **two hours**. Pages are independent and there is **no ordering constraint** here (unlike the taxonomy's grouping, where arrival order changes the result), so a bounded pool of 8 brings it to about **20 minutes**.

`NEED_EXTRACT_WORKERS` sets it, kept separate from the derivation's knob because the token profiles differ: a need response enumerates a page's tracked entries and is far larger than a referent list, so an operator may want the two rates set differently.

**The store stays inside the worker**, rather than collecting results for one write at the end. That's what makes an interrupted run resumable — every finished page is already durable, and the content-hash guard skips it next time. The taxonomy derivation wrote once at the end and lost 105 pages to a pod restart; this can't. A page that fails to store is logged and skipped rather than aborting the pool.

## Model

Needs now default to the **ingestion-pipeline model**, so they run on the same model the taxonomy they're labelled against was derived with (`gpt-5.6-luna` in prod).

Resolved to a **name**, never `None` or `""`: the staleness guard compares that column, so an empty value would make a later model switch invisible. Falls back to the main model's name when no ingestion model is nominated.

Worth flagging as a risk rather than hiding it: the earlier eval found weaker models did poorly on need extraction **for large pages specifically** — a stronger constraint here than for mention extraction, since `current_content` has to enumerate a page's entries rather than summarise them. Truncation now names the page, so if the ingestion model struggles we'll see it rather than infer it.

## Readiness check that prompted this

Audited `needs.py` against the failure modes the derivation hit. It already carried: output cap passed (16384), truncation surfaced, retry on malformed JSON, model recorded by name, prune scoped by prefix and guarded against an empty read. Sequential execution was the only gap.

Verified against prod: taxonomy `id=3` resolves, all 14 types splice into the prompt, task registered, 148 pages queued. One incidental find — the prompt is ~1,715 tokens, above OpenAI's 1024-token minimum shared prefix, so unlike entity extraction (~396 tokens, 0% cache hits) this should actually get cache discounts.

## Tests

6 new: pages extract concurrently; concurrency is bounded by the knob; every page is stored rather than just counted; one unstorable page doesn't abort the rest; the ingestion model is used; and it falls back to the main model *by name*.

Verified by mutation — forcing sequential, using the main model, or narrowing the per-page guard each fails a test. `ruff check` and basedpyright clean, full suite green.

### DIFF
--- a/backend/app/config.py
+++ b/backend/app/config.py
@@ -82,6 +82,10 @@ class Config(BaseModel):
     # call can carry ~47k input tokens, and this pool sits inside a queue task, so the real
     # concurrency is the queue's own worker count multiplied by this. Lower it on a lower tier.
     entity_type_derive_workers: int = 8
+    # How many need-extraction LLM calls run at once. Separate from the derivation's knob because
+    # the token profile differs: a need response enumerates a page's tracked entries, so it is far
+    # larger than a referent list, and an operator may want the two rates set differently.
+    need_extract_workers: int = 8
 
     # Relevance filter (drops irrelevant candidate pages before the LLM reconcile).
     # Empty model path => cosine cold-start filter; a path to an exported two-tower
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #592, opened 2026-08-03
fix(wiki): preserve list tightness through the markdown-Yjs round trip

Bo-Onyx:
## The asymmetry

The codec serialized every list loose (blank line between items) regardless of
how the source was written — documented as a rendering-only distinction. It is
content-preserving, but it means the file and the document permanently disagree
about every tight list's bytes: the first checkpoint that touches one rewrites
the whole list's spacing. In a co-edited page's history that reads as machine
commits nobody typed, and it erodes the byte-stable diff base the splice's
verbatim emission depends on.

This was found while root-causing today's task-list self-duplication incident:
the round trip `file → doc → file` was not the identity, and each touched-list
checkpoint committed the drift.

## The fix

- **Parse** records the distinction. markdown-it hides a tight item's paragraph
  tokens (CommonMark makes looseness a whole-list property, so per list they
  are all hidden or all visible); `_build_list` stamps `tight="true"` when
  hidden. Nested lists judge their own tokens at their own depth.
- **Serialize** joins a tight list's items — and each item's own blocks — with
  single newlines.
- **Safety valve** (`_tight_item_safe`): the attribute lives on the list node,
  so an editor mutation can hand the serializer a "tight" list whose item now
  holds a second paragraph; serializing that tightly would merge the paragraphs
  on reparse. Any block sequence that can't reparse tightly (paragraph after
  paragraph, `---` that would read as a setext heading, an ordered list not
  starting at 1, a table) drops the whole list to loose — one visible reflow,
  then stable.

Docs seeded before this change carry no attribute and keep the old loose
behaviour until their session reseeds — the safe direction.

## Verification

Replayed against the production page that surfaced it (a fresh-seeded doc,
touching each list block in turn, splicing against the committed body):

| touched block | before | after |
|---|---|---|
| tight 19-item task list | +17B (blank per item) | −2B (trailing-space trim) |
| tight 9-item task list | +8B | ±0 |
| loose task list | ±0 | ±0 |
| 2-item task list | +1B | ±0 |
| nested QA checklist | −57B | −128B (marker canonicalization, converges) |

Remaining diffs are one-time parse-level normalizations (trailing-space trims,
mixed-spacing lists normalizing to uniform loose per CommonMark, nested-marker
indent) — each converges in one pass and is byte-stable after.

New tests: tight/loose/nested byte-identical round trips, mixed-spacing
convergence, and the editor-added-second-paragraph fallback. Two existing tests
asserted the old lossy behaviour and were updated. Full backend suite green.

## What this does not fix

The duplication loop itself (a split task list fighting the id restamp) — that
is the companion frontend PR. This PR removes the formatting churn and restores
`file → doc → file` symmetry.

### DIFF
--- a/backend/app/wiki/markdown_yjs.py
+++ b/backend/app/wiki/markdown_yjs.py
@@ -692,6 +692,25 @@ def _build_list(
         if ordered:
             attrs["start"] = str(open_tok.attrs.get("start", 1))
 
+    # Tight vs. loose is recorded, not discarded: markdown-it hides a tight
+    # item's paragraph tokens (CommonMark makes looseness a whole-list
+    # property, so they're all hidden or all visible per list). The direct
+    # item paragraphs sit exactly two levels below this list's open token —
+    # a nested list's paragraphs are deeper and carry that list's own flag,
+    # judged by its own recursive ``_build_list`` call. Without this
+    # attribute every touched list re-serialized loose, so editing one item
+    # rewrote the whole list's spacing — phantom formatting churn in every
+    # co-edit checkpoint that touched a tight list.
+    para_level = open_tok.level + 2
+    direct_paras = [
+        t
+        for s, e in item_ranges
+        for t in tokens[s:e]
+        if t.type == "paragraph_open" and t.level == para_level
+    ]
+    if all(t.hidden for t in direct_paras):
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #593, opened 2026-08-03
fix(editor): join adjacent same-type lists; carry the codec's tight attr

Bo-Onyx:
Companion to #592 — together they restore the two directions of markdown↔doc
symmetry whose absence caused today's task-list self-duplication incident.

## The loop this kills

A document held two adjacent taskLists where the file's parse sees one (a list
split by an earlier edit). Reproduced byte-for-byte against production data
(the retained session snapshot replayed through the checkpoint splice — the
replay reproduces the real corrupt commit exactly, 10,853B, +324):

1. the checkpoint restamps block ids positionally against the file's parse →
   one file block covering two doc nodes gets **one id stamped onto both**
2. `UniqueBlockIdentity` clears the second node's id — its invariant
3. a cleared id reads as freshly-typed to `checkpoint_body` → it appends that
   node's serialization while the base's verbatim range still contains it →
   **+324B of duplicated items per checkpoint**
4. restamp re-shares, the clear re-fires — forever. 11 machine commits on the
   affected page today; the document itself stayed clean throughout, so the
   editor showed nothing wrong.

## The fix

`JoinAdjacentLists`: an `appendTransaction` that joins adjacent same-type
top-level lists (bullet/ordered/task). Markdown will reparse them as one list
regardless, so joining makes the document match every future parse of its own
file — the shape the restamp and the id check fight over stops existing.

- One join per pass; ProseMirror re-runs appended transactions, so longer runs
  converge without position remapping.
- The join keeps the first list's identity and counts as an edit, so the next
  checkpoint re-serializes the merged list from the document — **self-healing**:
  opening the corrupted page with this build wipes the file's duplicates on the
  next save.
- Ordered before `UniqueBlockIdentity`, which keeps its job for genuine
  one-node splits (paragraph Enter), where clearing the id is correct.

Also declares the `tight` list attribute (#592) in the schema so client-rebuilt
nodes don't drop it and reflow list spacing.

## Verification

Server-side arithmetic proven against prod data (see #592 and the incident
replay). The join transaction itself is typechecked but has no automated test —
the frontend has no test runner configured. I'd like to verify live on the
local stack (two tabs, split list, watch the join + the healing checkpoint)
before this deploys; happy to do that on request.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,6 +66,15 @@ repos:
         files: ^frontend/.*\.(ts|tsx)$
         pass_filenames: false
 
+      - id: vitest
+        name: vitest (frontend)
+        # Editor-behaviour tests (headless Tiptap) — sub-second suite, run
+        # whole on any frontend .ts/.tsx change like tsc above.
+        entry: bash -c 'cd frontend && bun run test'
+        language: system
+        files: ^frontend/.*\.(ts|tsx)$
+        pass_filenames: false
+
       - id: basedpyright
         name: basedpyright (backend)
         # basedpyright reads pyproject.toml for include/exclude — passing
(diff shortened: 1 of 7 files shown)


Bo-Onyx:
@greptile

PR #594, opened 2026-08-03
fix(editor): keep the doc text on the title's gutter at narrow widths

Bo-Onyx:
On a narrow window (especially with a side panel open), the doc body slid ~32px left of the page title, running toward the window edge while the title kept its inset.

Root cause: the editor's scroll wrapper is deliberately full-bleed — FileView cancels `<main>`'s padding with `-mx-8`/`-mx-3` so the scrollbar hugs the window edge — and defines `--cm-gutter` to mirror that padding so the text column keeps the title's side gutter. The old CodeMirror theme consumed that variable; the Tiptap migration's `.editor-prose .ProseMirror` column never did. While the width cap binds, centering happens to align both columns, so the bug only appears once the row gets narrow enough for the cap to stop binding.

Fix: `.ProseMirror` now consumes `--cm-gutter` as horizontal padding, with the cap grown by both gutters (border-box) so wide-layout text width is unchanged — narrow widths leave the text exactly on DocTitle's edge with the gutter as the minimum side space. The margin-comments-lane override (`.rail-reserved`, which re-declares the same geometry) grows by the gutters the same way, so reserving the lane no longer re-loses them.

Verified in the browser at 1130px with the Updates panel open (the reported repro): title and body share one left edge; wide layout unchanged.

### DIFF
--- a/frontend/src/app/css/editor.css
+++ b/frontend/src/app/css/editor.css
@@ -40,10 +40,19 @@
      edge) — matching DocTitle/the banners' width (--app-container-sm-md,
      Opal's sizes.css) so the title, banners, and doc text all share one
      column. The old CM6 editor capped this internally via its own theme;
-     ProseMirror has no equivalent, hence doing it here. */
+     ProseMirror has no equivalent, hence doing it here.
+
+     The wrapper is full-bleed (FileView cancels <main>'s padding so the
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #596, opened 2026-08-03
fix(editor): restore heading-bottom and list-row breathing room

Bo-Onyx:
A heading directly above a list sat flush on it (the Tailwind reset zeroes all margins and the editor only restores heading `margin-top`), and rows inside a tight list rendered with exactly 0px between them.

Scanned before styling: the codec models every authored blank line as an empty paragraph (`markdown_yjs` — an inter-block blank line is a first-class 24px node), so the doc's larger rhythm is already the author's blank lines. Standard prose margins (like the read-view `.markdown` rules) would double-space every existing page on top of those. Verified against the live DOM: computed margins on `ul`/`li`/`p` are all 0 and the item-to-item gap is 0px.

So the change is deliberately minimal:
- `h1`–`h6` get `margin-bottom: 0.25em` — fixes the reported heading-hugs-list case;
- `li` gets 2px top/bottom — a tight list reads as rows, not a brick, without touching the blank-line rhythm.

Verified in the browser on a heading + tight-bullet page: clear gap under the heading, small even row gaps, blank-line-spaced content unchanged.
Before
Screenshot 2026-08-03 at 4 42 31 PM
after
Screenshot 2026-08-03 at 4 43 02 PM

### DIFF
--- a/frontend/src/app/css/editor.css
+++ b/frontend/src/app/css/editor.css
@@ -94,12 +94,11 @@
   color: var(--text-05);
 }
 
-/* Breathing room the Tailwind reset removed. Kept deliberately small: the
-   doc's larger rhythm is the author's own blank lines — the codec models
-   each one as an empty paragraph (see markdown_yjs) — so real block
-   margins here would double-space every existing page. Headings get a
-   bottom margin (a heading directly above a list sat flush on it), list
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
@greptile

PR #597, opened 2026-08-03
fix(coedit): stop the duplicate-id repair from eating a split's new half

Bo-Onyx:
Caught **live during end-to-end verification** of #592/#593 on a local stack:
lift a paragraph out of a list, type into it, and the next autosave deleted the
typed text. Worker log: `held duplicate top-level block ids ['b4']; dropped the
repeats`.

## The race

1. ProseMirror's node split copies attrs — `_blockId` included — onto both
   halves, so the freshly created block briefly shares the original's id.
2. The editor's `UniqueBlockIdentity` clears the second id — but that is a
   **separate follow-up update**.
3. An autosave checkpoint landing between the two sees a repeated id.
   `drop_duplicate_blocks` (from #585) assumed repeated id ⇒ lineage merge ⇒
   copies identical ⇒ delete the later one. Here the later one is the
   person's edit. Deleted, broadcast, gone.

The #585 docstring even asserted "freshly typed content carries no id at all"
— false for split-created blocks, for exactly the window before the client's
fix-up lands.

## The fix

Distinguish by content, repair accordingly:

- **identical copy** → still deleted (the lineage-merge case #585 targeted;
  broadcast-converged as before)
- **differing copy** → its id is **cleared** instead — the same repair the
  client's own plugin was about to apply — routing it through
  `checkpoint_body`'s `orig is None` path as the new content it is. Nothing
  lost, id invariant restored either way.

## Testing

New test builds the exact race shape (a paragraph carrying a list's copied id,
with typed text) and asserts both blocks survive the repair and the splice.
The pre-existing lineage-merge tests pass unchanged. Full backend suite green
(2404 passed).

Live re-verification on the local stack follows once this lands in the image.

### DIFF
--- a/backend/app/wiki/coedit_checkpoint.py
+++ b/backend/app/wiki/coedit_checkpoint.py
@@ -101,35 +101,54 @@ def drop_duplicate_blocks(doc: Doc) -> list[str]:
     client holds, the client's document converges to the repaired state
     without a reload (verified directly against pycrdt).
 
-    A repeat with **different content** is an ordinary edit caught
-    mid-flight: ProseMirror's node split copies attrs — the id included —
-    onto both halves, and the editor's ``UniqueBlockIdentity`` clears the
-    second one in a separate follow-up update. A checkpoint landing between
-    the two sees the duplicate the client is about to fix. Deleting here
-    destroys what the person just typed (observed live: a paragraph lifted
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #599, opened 2026-08-04
feat(ingest): capture each need's stated update instruction

Bo-Onyx:
A page often writes down **how it is to be maintained**, and nothing in the need schema held that.

`detail_level` is *inferred* from the entries already present. An update instruction is a directive the author **wrote**, and it constrains things no amount of reading content reveals. Three real examples from this wiki:

| kind | instruction |
|---|---|
| **placement** | *"Add each Friday's notes as a new dated section below, newest first."* |
| **admission** | *"all information should come here either through the customer, or from internal sources explicitly referencing information provided by the customer. Overall deal status should come from a CRM system."* |
| **format** | *"Each line should include the source type where it came from, which customer it is (or Unknown), the date, and what the issue is."* |

The **admission** rule is the argument for the field: it decides whether an incoming document is eligible at all — something `detail_level` cannot express, and exactly the judgment a reconciler needs before it edits a page.

Worth noting the eval saw these only **incidentally**, in a free-text `evidence` field on 12 of 205 needs. The information was visible and had nowhere to live.

## Two deliberate constraints

**Verbatim, not paraphrased.** A reworded directive can permit or forbid something the original didn't. This field is meant to be authoritative, so it quotes.

**Empty when absent, and the prompt says so twice.** Most pages state no rule. A fabricated instruction would be obeyed as though a human wrote it — worse than having none. Same reasoning that led to dropping organization-name detection: being wrong here is silent and compounds, while being absent costs nothing.

Also scoped to rules about *maintaining the page*: a runbook's steps or a design doc's guidance to engineers are content, not update rules.

## Re-extraction required

This makes every stored need stale, and the prompt is deliberately **not** part of the staleness guard (stored needs are only comparable when they came from the same prompt). So the re-run needs `force=True` — 148 pages, ~6 min on the ingestion model.

## Tests

5 new: the instruction is captured; absence stays absent across `""`/`None`/whitespace; it's independent of `detail_level`; it reaches stored JSON; and the prompt asks for it verbatim *and* warns against inventing one.

Verified by mutation — dropping the field, or softening the do-not-invent rule, each fails a test. `ruff check` and basedpyright clean, full suite green.

## Context

Prior run for comparison, on the current schema: 255 needs from 148 pages in 6m06s, 2,903 entities all typed against taxonomy id=3, zero truncations, `current_content` enumerating real values rather than describing pages. This adds one field to that.

### DIFF
--- a/backend/app/db/models.py
+++ b/backend/app/db/models.py
@@ -2102,7 +2102,7 @@ class PageNeeds(Base):
     entity_type_taxonomy_id: Mapped[int | None] = mapped_column(
         Integer, ForeignKey("entity_type_taxonomies.id", ondelete="SET NULL")
     )
-    # [{need_name, need_kind, description, detail_level, current_content,
+    # [{need_name, need_kind, description, detail_level, update_instruction, current_content,
     #   entities: [{canonical_name, entity_type, primary}], focus}, ...]
     # JSONB and free-form for the same reason as ``EntityTypeTaxonomy.types``: the shape is owned
     # by ``app.ingest.needs``, read whole, and never queried by field.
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
@greptile

PR #600, opened 2026-08-04
fix(coedit): reconnect through join failures; tell the user what actually helps

Bo-Onyx:
## The report

The indicator sometimes shows **"Couldn't save: connection closed"** and it's
unclear what the user should do — wait? refresh?

Tracing the two paths behind it:

1. **Socket dropped with a save in flight.** Self-healing: the loop reconnects
   on the same document every 3s, disconnected edits merge on reconnect, the
   save retries. The message reads as a verdict anyway — and the natural
   response, *refreshing*, is the one action that discards unsynced edits.
2. **A reconnect attempt fails at the join handshake** (e.g. it lands inside a
   backend deploy's seconds of downtime). The loop **gave up permanently**,
   stranding the tab behind a "Couldn't connect / Retry" banner for an outage
   that had already passed — and Retry rebuilds the document, discarding
   unsynced edits, same as a refresh.

## Changes

- **Reconnect through join failures.** A failed attempt backs off (3s doubling
  to a 30s ceiling) and retries on the *same* document, indefinitely. Only the
  genuinely non-retryable join — the codec-unsupported page — still ends the
  loop, so `joinError` is now terminal-only and the dead-end Retry banner no
  longer appears for transient outages.
- **One truthful state for the whole outage**, rendered with precedence over
  save errors (which are consequences of the drop): "Reconnecting…", and after
  four attempts: *"Still reconnecting — keep this tab open; edits save
  automatically once back."* The tooltip carries the full explanation. Keeping
  the tab open is the only instruction that matters — it's the action that
  preserves the unsynced edits.
- **The promise is made true**: a successful reconnect re-fires the save, so a
  retry budget exhausted mid-outage can't leave a stale failure that only more
  typing would clear. A clean session makes the kick a server-side no-op.
- Save rejections that merely restate the outage ("connection closed" / "not
  connected") now carry instruction — *"saves automatically once reconnected —
  keep this tab open"* — instead of a raw transport diagnosis.

## Testing

`tsc`, prettier, and the editor Vitest suite pass. The reconnect loop's
behaviour is hook-level (WS + timers); verified by reading, not by an
automated test — a jsdom harness for the hook is the existing open "editor
smoke test" item. Happy to run a live check (kill the backend mid-edit, watch
the states) on the local stack before this deploys.

### DIFF
--- a/frontend/src/lib/editor/hooks.ts
+++ b/frontend/src/lib/editor/hooks.ts
@@ -304,6 +304,8 @@ export function useCoeditSession(opts: {
     useState<UseCoeditSession["saveStatus"]>("saved");
   const [saveError, setSaveError] = useState<string | null>(null);
   const saveRetryTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  // The post-reconnect save kick, cancellable on teardown/reconnect.
+  const reconnectKick = useRef<ReturnType<typeof setTimeout> | null>(null);
   const saveRetries = useRef(0);
   // Counted apart from saveRetries: a timeout and a "not connected" cost two
   // very different amounts of wall-clock, so they can't share a budget.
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #602, opened 2026-08-04
fix(wiki): path-free suggestion titles; clamp the card's rows to its width

Bo-Onyx:
## The report

The Auto-Organize suggestions card rendered badly with real paths: row titles
ran underneath the per-row approve/reject buttons and pushed the card wider
than the side panel, horizontally scrolling the whole panel.

## Root cause

Two stacked problems:

- The row title was `proposal.summary`, which quotes the **full wiki path** —
  redundant with the folder tag rendered directly below it.
- The content column uses `alignItems="start"`, which in a flex column sizes
  children to their **content width** — so the title's `maxLines={1}`
  truncation and the tag never received a width constraint. The text rendered
  at full intrinsic width, under the buttons and past the card edge.

## The fix (per the design guidance)

- **Titles are path-free**, derived from the proposal op: "Remove empty
  folder", "Remove page", "Move page", … The tag below carries the path.
  Unknown ops fall back to the summary.
- **Longer titles wrap** inside the card instead of truncating — made possible
  by `w-full min-w-0` wrappers on the title and tag (without the constraint,
  wrapping can't happen either).
- The **full summary rides the row's hover tooltip**, so the detail (e.g.
  *"— still identical to the template it was created from"*) stays reachable.
- The path tag uses Opal `Tag`'s own `truncate` prop.

## Verified

Live on a local stack with two seeded proposals carrying long paths: rows read
"Remove empty folder" / "Remove page", tags truncate with an ellipsis, buttons
sit clear of the text, no horizontal overflow. `tsc` + prettier clean.

Screenshot 2026-08-04 at 10 33 54 AM

### DIFF
--- a/frontend/src/components/wiki/SuggestionsCard.tsx
+++ b/frontend/src/components/wiki/SuggestionsCard.tsx
@@ -17,6 +17,7 @@ import {
   SvgFiles,
   SvgFold,
   SvgFolder,
+  SvgArrowRight,
   SvgFolderIn,
   SvgFolderPlus,
   SvgSettings,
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
@greptile

PR #605, opened 2026-08-04
feat(ingest): cluster needs into the facets that span pages

Bo-Onyx:
Need extraction is per page, so it can't see that nineteen pages all track implementation status. This is the step that can: embed every need, cluster the embeddings, and the clusters reaching more than one page are the shared facets.

That reach is the premise the whole layer rests on — a fact belonging to a shared facet can be reconciled **once** and applied to every page holding it, instead of each page independently deciding what to do with the same document.

## What it measures on real data

238 needs from a 149-page production wiki:

| | |
|---|---|
| clusters | 156 |
| **spanning >1 page** | **26** |
| needs in those | 99 (42%) |
| widest reach | 24 pages |

So the fan-out is real — and so is the half that's genuinely page-local. A clustering that claimed everything was shared would be wrong.

Worth seeing the distribution, because it's more concentrated than the headline suggests: `24, 7, 6, 4, 4, 3, 3, 3` then eighteen 2s. One genuine hub, then mostly pairs.

## The embed key, and a mistake worth recording

`need_name` + `description`, at **0.60**. Both settled against two labels held **out** of the key — a need's kind, and its primary entity's type — on the principle that a real facet cluster should agree with signals it never saw. At matched cluster counts: **71% / 73%**, against 68% / 65% for name-only.

I first concluded description was noise. That was wrong, and the reason is worth stating: **comparing keys at a fixed threshold is a trap.** Adding description raises every pairwise similarity, so the same number buys coarser clusters and looks worse for reasons unrelated to the key's quality. Compare at matched *granularity* instead.

What's excluded, and why:

- `current_content` — the **state**, which churns on every edit. In the key, a need would move between clusters when its content changed rather than when what it tracks changed.
- `need_kind` — a closed four-value vocabulary. Appending it pulled 113 "reference" needs together into a **70-page cluster**: grouping by kind, not subject.
- `entities` — the other axis. An entity is the **row**, a facet is the **column**; embedding entities clusters "everything about Scania" instead of "deal status, across customers", collapsing exactly the fan-out this exists to find.

## Grouping only

Naming a cluster's topic is an LLM step and lives apart, so this stays deterministic, inspectable, and reproducible — which a naming step downstream depends on.

**Known limitation, stated rather than hidden:** I read all 26 multi-page clusters. 24 are coherent (one is *weekly customer progress / GTM review / all-hands / team standup* across 7 pages — a real facet). **Two are junk drawers**, the largest holding 25 needs across 24 pages joined only by being *about Agent Wiki* — a subject, not a facet. That's centroid drift, and raising the threshold breaks the good pairs before it fixes them (at 0.65 the widest only drops 24→13 while multi-page clusters fall 26→15). The designed answer is the partition step in naming, which splits a mixed cluster — but that's an assumption I'd rather test than assert.

## Also here

`app/ingest/clustering.py` — the vector helpers extracted from `entity_types` now that this is a second consumer, so neither module imports the other's privates. No behaviour change to entity-type derivation.

## Tests

16 new: the key includes name+description and excludes state/kind/entities; similar needs across pages group; a single-page cluster isn't a failure; **unavailable embeddings return `None` rather than singletons** (otherwise "no embeddings" is indistinguishable from "nothing is shared"); clusters order by reach; the same input clusters the same way twice; disabled pages are excluded; plus the helpers.

Mutation-verified. `ruff check` and basedpyright clean, full suite green.

Split from the policy-filter work, which is now its own PR and independent of this.

### DIFF
--- a/backend/app/ingest/clustering.py
+++ b/backend/app/ingest/clustering.py
@@ -1,0 +1,56 @@
+"""Vector helpers shared by the derivation stages that group things by similarity.
+
+Extracted from ``app.ingest.entity_types`` when need clustering became a second consumer. The
+functions are unchanged; only the home is new, so both callers use one implementation rather than
+one importing the other's privates.
+"""
+
+from __future__ import annotations
+
+import math
+
+
+def normalize(vec: list[float]) -> list[float]:
+    norm = math.sqrt(sum(v * v for v in vec)) or 1.0
+    return [v / norm for v in vec]
+
+
+def cosine(a: list[float], b: list[float]) -> float:
+    """Both operands are unit vectors, so the dot product IS the cosine."""
+    return sum(x * y for x, y in zip(a, b))
+
+
+def leader_cluster(
+    vectors: list[list[float]], order: list[int], threshold: float
+) -> list[list[int]]:
+    """Greedy leader clustering: each item joins the first cluster whose centroid is within
+    ``threshold``, else seeds a new one.
+
+    Chosen over agglomerative linkage because the backend has no scipy/sklearn and an O(n^3)
+    pure-Python linkage would not finish on a real corpus. ``order`` should put the
+    best-supported items first so they seed clusters — that makes the result deterministic and
+    puts the strongest evidence in charge of each group.
+
+    A threshold tuned for this algorithm does NOT transfer to a linkage method: the centroid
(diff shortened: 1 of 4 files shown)


Bo-Onyx:
@greptile

PR #606, opened 2026-08-04
feat(ingest): skip auto-update-disabled pages; carry the policy instruction

Bo-Onyx:
Needs are only useful where the ingestion pipeline may write. A need on a page that can never be auto-updated isn't merely wasted spend — it **presents itself as somewhere a fact could be reconciled to**, when nothing may write there.

On the production wiki that's **37 of 149 pages (25%)**, holding **66 of 238 stored needs (28%)**.

## The filter

`update_policy.resolve_for_paths` handles the tri-state through the folder cascade and applies the `ai_edits_disabled` master override, so this is one call rather than a rule reimplemented here.

Retiring is free: `prune` is driven by the live set, so a page turned off *after* extraction drops out with no separate cleanup path.

## The policy instruction

Each need now carries the **admin's configured** `update_instruction` alongside the one **extracted from the page's own text** — one a human's directive, one what the page tells its readers.

You asked whether a dedicated field earns its place given `description`/`detail_level` exist. Measured over the 238 stored needs: the page-stated instruction shares a **median 17%** of its words with those fields, and **51 of 61 share under 40%**:

```
[0%] do not re-add module-level `depends_on` to modules containing data sources.
[0%] Revisit if multi-client demand shows up; don't front-load it.
[0%] Steps use checkbox (`- [ ]`) syntax for tracking.
[0%] Update at checkpoints.
```

Prohibitions, deferrals, syntax requirements, cadence — `description` says *what* a page tracks and `detail_level` *how granular*; neither can say "do not re-add X".

**The policy rule is kept out of the prompt on purpose.** Shown the admin's rule, the model would echo it back as the page's own instruction and destroy the verbatim-from-page property that makes that field trustworthy.

**And it's in the staleness guard.** Policy is governance state — an admin can rewrite it with the page untouched. `content_sha256` now hashes it alongside the body, so editing a rule re-extracts the pages it governs. Without that, needs derived under a superseded rule stay stored looking current forever — the failure the taxonomy id already prevents.

## Tests

5 new: disabled pages aren't extracted; disabling retires existing needs; the policy instruction is stored per need; editing it re-extracts; an unchanged policy doesn't.

`ruff check` and basedpyright clean, full suite green.

Split out of #605; the clustering half is now #605 and the two are independent.

### DIFF
--- a/backend/app/db/page_needs.py
+++ b/backend/app/db/page_needs.py
@@ -52,14 +52,31 @@ def _now() -> str:
     return datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M:%S")
 
 
-def content_sha256(body: str) -> str:
-    """Hash of a page body — the re-extract guard. Uncapped, unlike the embedding store's:
-    extraction sends the page whole, so a change past any cap still changes the result."""
-    return hashlib.sha256(body.encode("utf-8")).hexdigest()
+def content_sha256(body: str, policy_instruction: str = "") -> str:
+    """Hash of everything a page's needs are derived FROM — the re-extract guard.
+
+    Uncapped, unlike the embedding store's: extraction sends the page whole, so a change past any
+    cap still changes the result.
+
+    ``policy_instruction`` is hashed alongside the body because it is an INPUT to extraction, not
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

PR #607, opened 2026-08-04
fix(wiki): page proposals move to the Updates panel; retire the top banner

Bo-Onyx:
Aligned direction: proposals surface in one place — the Auto-Organize
suggestions card in the right-hand **Updates** panel. Folder views already do
this; page views still mounted the older `Path2ReviewBanner` (#478) above the
document body, pushing content down and duplicating the pattern the card
replaced.

## Changes

- Page views host `SuggestionsCard` in their Updates tab, directly beneath the
  policy cards — the same arrangement as the folder side panel. So a
  page-scoped proposal (e.g. stale-page delete) is reviewed where every other
  proposal is.
- The Updates tab's `UpdatePolicyPanel` drops `fullHeight` so the stack sizes
  to content — otherwise the card pinned to the panel bottom past a stretch
  gap (observed live before the fix).
- `Path2ReviewBanner` had no remaining mounts and is **deleted**.

## Verified live (local stack)

Seeded a page-scoped stale-page proposal: the page body renders clean with no
banner, and the Updates panel shows settings → instructions → history → the
suggestions card with the proposal ("Remove page", path tag, approve/reject).
Screenshot 2026-08-04 at 12 31 35 PM
Screenshot 2026-08-04 at 12 31 26 PM
Screenshot 2026-08-04 at 12 31 23 PM
Screenshot 2026-08-04 at 12 30 59 PM
Screenshot 2026-08-04 at 12 30 56 PM


Bo-Onyx:
Extended per the aligned design (second commit): retiring the banner alone left page proposals discoverable only behind the panel toggle — the intended behaviour is the folder view's floating pattern, now ported to pages:

- **Self-showing**: with pending proposals (or an update-frequency warning), the notification cards float under the header's Auto icon on page load — no hover needed. X hides them for the page; an open updates panel supersedes them.
- **Hover stacking**: hovering the Auto icon stacks the AI Auto-Edits settings **above** the notification cards, pushing them down — mock 2283:84706, same as folders.
- The update-frequency warning moved into the same stack; its above-the-document banner mount is gone, so nothing pushes page content down anymore.

Both behaviours live-verified on a local stack (screenshots in the session). Known parity gap, pre-existing: `PresenceAvatars` is desktop-only, so mobile has neither surface — same gap already tracked for agent presence chips.

PR #608, opened 2026-08-04
feat(health): queues describe themselves for the admin status page

Bo-Onyx:
The admin Status page labels queues from a frontend map that has drifted: it has three entries, so the three queues added since (`coedit`, `automanage_offline`, `automanage_nearline`) render as raw internal ids.

This moves the display contract to the source: `TaskQueue` gains `label` and `description` fields, each queue defines its own beside its registration in `app/tasks/queues.py` (the description says what a backlog means for a person — "a human is waiting" vs "harmless"), and `GET /health` serves them per queue. A future queue can't silently regress the page: the health test asserts every queue ships a non-empty label and description.

Additive response change — the current frontend ignores the new fields. The frontend PR that renders them follows separately.

Labels shipped: Document updates / Watcher evaluations / Editor saves / Auto Organize sweeps / Auto Organize approvals / Indexing & cleanup.

### DIFF
--- a/backend/app/api/health.py
+++ b/backend/app/api/health.py
@@ -36,6 +36,8 @@ def _build() -> HealthResponse:
             error = str(e)
         queues.append(QueueHealth(
             name=name,
+            label=queue.label or name,
+            description=queue.description,
             ready=ready,
             delayed=delayed,
             in_flight=in_flight,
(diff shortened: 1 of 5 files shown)


Bo-Onyx:
@greptile

PR #609, opened 2026-08-04
feat(health): render served queue labels and backlog descriptions

Bo-Onyx:
Frontend half of #608. The Status page's queue cards now render the label and what-a-backlog-means line each queue serves about itself, replacing the local `QUEUE_LABELS` map that had drifted (three entries; `coedit` and both `automanage` queues rendered as raw ids). Stat wording becomes person-first: `waiting / scheduled / running` (with title tooltips) instead of `ready / scheduled / in flight`.

The new response fields are optional in the client type and the card falls back to the queue name, so this renders correctly against a backend that predates #608 — deploy order doesn't matter.
Screenshot 2026-08-04 at 1 27 15 PM

### DIFF
--- a/frontend/src/app/admin/health/page.tsx
+++ b/frontend/src/app/admin/health/page.tsx
@@ -10,12 +10,6 @@ import { useIsMobile } from "@/lib/viewport";
 
 const POLL_MS = 5000;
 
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #610, opened 2026-08-04
fix(wiki): applied changes reach every open view

Bo-Onyx:
## Recovery note

#607 was merged while its last two commits were still in flight — the squash
contains only the first two. This PR carries the orphaned pair verbatim
(cherry-picked from the closed branch) plus the new staleness fix:

1. **Row click navigates to the affected page/folder** (was `3fadc40a`) — a
   proposal can target something nested below the current view, and deciding
   on a proposal usually starts with looking at the thing.
2. **Background revalidation + reason in the title** (was `af126385`) — the
   proposals list refreshes every 30s and on focus so another reviewer's
   action reaches open views (acted rows kept locally so outcomes stay
   visible), and the summary's reason clause rides the row title, wrapping
   over lines.

## New: applied changes reach every open view

The reported staleness: someone applies a change and a colleague's focused tab
never updates — tree and listings only revalidated on focus. Three layers:

- the shared `/wiki` listing **polls every 60s** (structure changes constantly
  under focused tabs — agents, ingestion, colleagues; page bodies are already
  live via the co-edit session, so structure was the gap);
- **the actor refreshes instantly**: an "applied" outcome triggers
  `revalidateWiki()` the moment it lands;
- **viewers follow within the proposals poll**: proposals vanishing without
  local action (someone else acted) trigger the same refresh.

`tsc`, prettier, editor Vitest suite green. Live-verified on a local stack.

@nikolas-garza same surface as #602/#607.

### DIFF
--- a/frontend/src/components/wiki/SuggestionsCard.tsx
+++ b/frontend/src/components/wiki/SuggestionsCard.tsx
@@ -71,6 +71,18 @@ function displayPath(path: string, budget = 28): string {
   return head ? `${head}/…/${leaf}` : `…/${leaf}`;
 }
 
+/** The reason clause of a backend summary — the text after the " — " that
+ * follows the quoted path (`Remove “x.md” — still identical to …`). The
+ * first closing quote ends the path, so a quote inside the reason itself
+ * doesn't confuse the split; summaries with no quoted path fall back to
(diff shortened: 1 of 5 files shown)


Bo-Onyx:
@greptile

PR #612, opened 2026-08-04
feat(db): need map schema — topics, aspects, and the needs composing them

Bo-Onyx:
Schema only. Landing the shape first because a migration is the expensive thing to get wrong — and this one changed three times while the PR was open, each time because a field couldn't justify itself.

```
need_map    one derivation
  topic       the SUBJECT   "Wiki Auto Management"
    aspect      a FACET     "implementation status"
      needs       what composes it, on their pages   ← the fan-out
```

Fan-out is a property of the **aspect**, not the topic: judge a fact once against an aspect and it applies to every page under it. Not literally one output for *n* pages — pages state their own granularity — so it is **decide once, render per page**.

## Scope

| | |
|---|---|
| **in** | tables, migration, read/write store, tests |
| **out** | the producer that fills them, aspect **state**, reconciler changes |

Nothing consumes this yet. It is deliberately inert.

## Named for its input

`need_maps`, not `topic_maps` or `topic_map_runs`.

A derivation's only genuinely new information is **a grouping of the corpus's needs, plus names for the groups**. Everything else in these tables is a pointer. So the name follows the input:

- `topic_maps` privileged the decorative half — nothing downstream reads the topic, because reconciliation happens at the aspect, which carries its own pages
- `_runs` described a job execution rather than the artifact
- `corpus_fingerprint` fingerprints **needs**, which is the giveaway: a map is stale when its input moves

`need_map_id` also reads better on `topics` and `aspects` than `run_id`, which begs "run of what".

## Deliberately not here: state

The current value per aspect — what an incoming fact gets diffed against — is **absent**, and should be. It changes with every relevant document while these tables are a snapshot.

That matches where Joachim landed independently in the eval repo: `meta_topics.json` and `meta_topic_statuses.json` are separate artifacts, because *"status is volatile (updated by every relevant doc), needs are stable (re-derived from wikis rarely)"*. His status is `status[aspect_id][entity_key] → {public, private}` — per aspect, then per entity, so two customers' deal values never commingle.

**Two things follow, and they change the build order rather than this PR:**

**Aspect identity has to be solved before state.** Ids are stable only *within* a map, so the first re-derivation orphans every state cell — and state is the expensive half, since it accumulates from every document ever ingested. Re-seeding from the pages recovers what they currently say and discards everything the document side contributed.

**State and structure want to come from one call.** In S3 the same LLM call that groups needs into aspects also reconciles their `current_content` into one merged status — because reconciling state *needs* the grouping and all the member content, which that call already has. So state should land before or with the producer, not after it, or we pay that call twice.

## Four tables, not one JSONB document

**A page reference is a real foreign key.** A deleted page can't leave a row pointing at nothing; inside a blob that would need filtering at every read. Pinned by a test that fails when the property is removed.

An earlier revision had five, on a second argument: that an aspect belongs to more than one topic. **That was my assertion, never an observation.** A second topic link changes no behaviour — reconciliation happens at the aspect, reached through `aspects_for_page()`, and the topic is never consulted. No producer can emit a shared aspect either: any pipeline that partitions a cluster into topics and then groups within a topic yields topic-local aspects by construction. So `topic_aspects` went and `aspects` gained `topic_id`.

If topics later become a *retrieval* unit, restoring the join is a backfill from that column — about the cost of removing it. Symmetric, so the simpler shape wins until there's evidence.

Scale agrees rather than conflicts:

| corpus | as a document | as tables |
|---|---|---|
| 149 pages | 0.1 MB | 228 rows |
| 10,000 pages | **8.5 MB** | 15k rows |

8.5 MB is fine to cache, wrong to read per incoming document — while `aspects_for_page()` stays an indexed lookup at any size.

## The map stores structure; needs stay where they live

`aspect_pages` is three columns — `aspect_id, doc_id, need_name` — and carries no copy of the need.

It briefly carried five more. The motivation was real: needs have no table of their own (a JSONB list on `page_needs`), so there is nothing to JOIN to, and copying made `aspects_for_page()` self-sufficient. The cost was worse than the saving, because `page_needs` is **current-valued** and this map is a **snapshot**:

| | when the page changes |
|---|---|
| pointer only | read-through finds nothing → no kind → reconciler infers from the body, which it can already do |
| copies | copy is still there, silently wrong → appends where it should replace |

Degrading to *no information* is recoverable; degrading to *wrong information* isn't, and is invisible. Structure is also the stable half — a page edit rarely changes which pages share a facet, often changes how that page maintains it.

If read-through later measures expensive, the answer is a **cache with its guard** (copies plus the `content_sha256` they were read at), not bare copies. Not now: nothing consumes the map, so there is no cost to optimise against.

## A dropped-need bug, fixed by the key

The old key was `(aspect_id, doc_id, entity)` — the page as the unit. But clustering can put **two of one page's needs in one aspect** (`test_a_single_page_cluster_is_not_a_failure` does exactly that); both carry `entity = ""`, so `record()` discarded the second with a debug log and the primary key would have rejected it anyway.

The key is now `(aspect_id, doc_id, need_name)` — the **need** is the unit. Mutation-verified: restoring the page-only key fails both tests.

`entity` isn't modelled here at all. A need's entity list is the need's own, and — per Joachim's status — entity belongs in *state*, not structure.

## Verified before designing it

One partition call over the widest real cluster (19 needs / 18 pages, joined only by being *about Agent Wiki*) returned **11 topics with 19/19 coverage**, including exactly this shape:

```
Wiki Auto Management
  requirements and rollout   → 1 page
  architecture and design    → 1 page
  implementation status      → 2 pages     ← fan-out the embedding had buried
```

That settled an assumption I'd flagged rather than asserted: the naming pass *does* rescue a junk-drawer cluster instead of naming it something vacuous. The **aspect-grouping half is still untested** — it runs for the first time in the producer.

## Map semantics

Everything is scoped to a map and cascades from it, so a derivation is one insert and a prune is one delete.

Written **whole, never updated in place** — clustering is global, so one need changing can move cluster membership anywhere, and revising a single topic isn't a coherent operation. Superseded maps stay readable; ids are stable only *within* a map.

`corpus_fingerprint` is built from each need set's own guard hash, not page bodies — so a page edit that leaves its needs unchanged doesn't make the map look stale. That matters because the naming step is a sampled call: re-deriving on unchanged needs would churn topic names for nothing, the same instability seen with the taxonomy (18 → 23 → 14 types on identical input).

**Known gap:** nothing reads `corpus_fingerprint` yet. It's computed, stored, and never compared — the guard should be wired up by whatever first consumes the map.

## Known-weak, stated rather than hidden

`need_name` is the pointer into `page_needs`, and it is LLM-generated per extraction. A surrogate id would not fix that — re-extraction regenerates needs wholesale, so keeping ids across it needs the same name-matching, and an FK with CASCADE would silently delete map links on every page edit. The prerequisite is **need continuity** (match new needs to old, preserve identity), not a table.

Two cheap follow-ups, neither in this PR: a bulk `page_needs.get_by_doc_ids()` so a fan-out is one query, and a count of `aspect_pages` rows that fail to resolve — which turns invisible decay into the number that decides whether continuity is worth building.

## Aligned with the eval's schema where it earns it

**Taken:** a one-line statement of what a subject *is* — a thin aspect name like "implementation status" is ambiguous alone. Stored as `description` rather than `gist`.

**Not carried back — `subject_entity_type`:** no downstream reads it, and the rule that would fill it was wrong more often than right on this corpus.

**Not carried back — `cadence`:** measured here and dropped — populated on 9–18% of needs, missing from most timelines under a weaker model, restating `detail_level` when present.

## Tests

**21** in `test_need_map_store.py`, plus **3** in `test_migration_chain.py`. 222 green across the migration, model, schema, topic and needs selection; `ruff` and basedpyright clean.

The migration test earns its place: `upgrade()` short-circuits when the tables exist, and fresh databases are built from the models — so the body is dead code on a normal run and a drifted `create_table` would be invisible. The round-trip rewinds past it, upgrades back, and asserts the resulting columns and primary key. It is also the only thing exercising the downgrade's drop order, which has broken twice.

Rebased onto main, which added `c5d8e2f7a941` revising this migration's parent — re-parented onto `b3d9f4c07a21` so alembic keeps a single head.

## Open

- **Aspect identity across derivations** — blocks state. Likely a stable natural key over `(topic name, aspect name)`.
- **Whether kind-mixed aspects are real or a clustering artifact.** Production has one "implementation status" kept as a dated log on one page and a current-state checklist on another. If that's common, per-page kind is right; if rare, it's a merge the embedding shouldn't have made and clustering should split on it. Needs one read-only export and ~238 embeddings; it shapes the producer's prompt, so it's worth doing before that.

### DIFF
--- a/backend/app/db/migrations/versions/a4d92e1c7f38_need_maps.py
+++ b/backend/app/db/migrations/versions/a4d92e1c7f38_need_maps.py
@@ -1,18 +1,23 @@
-"""topic map — runs, topics, aspects, and the pages holding them
+"""need map — one derivation of topics, aspects, and the needs composing them
 
-The derived topic layer: subjects, their facets, and which pages hold each facet.
+The derived layer over page_needs: subjects, their facets, and which needs compose each
+facet. Named for its input, not its output — a derivation groups the corpus's NEEDS, and the
+topics and aspects are the names it invents for those groups.
 
 Relational rather than one JSONB document because a page reference is a real foreign key: a
 deleted page cannot leave a row pointing at nothing, which inside a blob would need filtering at
 every read. Size is not the argument — at ten thousand pages the equivalent document is ~8.5 MB,
 fine to cache but wrong to read per incoming document, while the reverse lookup stays indexed.
 
-Everything is scoped to a run and cascades from it, so a derivation is one insert and a prune is
-one delete. One run is ``active``; the rest stay readable, because topic and aspect ids are only
-stable WITHIN a run.
+Everything is scoped to a map and cascades from it, so a derivation is one insert and a prune is
+one delete. One map is ``active``; the rest stay readable, because topic and aspect ids are only
+stable WITHIN a map.
+
+State — the current value per aspect — is deliberately NOT here. It changes with every relevant
+document while this is a snapshot, so it belongs in its own current-valued table.
 
 Revision ID: a4d92e1c7f38
-Revises: f2c9a41e7b06
+Revises: b3d9f4c07a21
 Create Date: 2026-08-04 20:40:00.000000+00:00
 """
 
@@ -25,7 +30,7 @@ from alembic import op
 from sqlalchemy.dialects import postgresql
 
 revision: str = "a4d92e1c7f38"
-down_revision: str | None = "f2c9a41e7b06"
+down_revision: str | None = "b3d9f4c07a21"
 branch_labels: str | Sequence[str] | None = None
 depends_on: str | Sequence[str] | None = None
 
@@ -36,11 +41,11 @@ def upgrade() -> None:
     # Guarded because ``0001_initial`` builds fresh databases from the current models, which
     # already carry these tables.
     inspector = sa.inspect(op.get_bind())
-    if inspector.has_table("topic_map_runs"):
+    if inspector.has_table("need_maps"):
         return
 
     op.create_table(
-        "topic_map_runs",
+        "need_maps",
         sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
         sa.Column("active", sa.Boolean(), server_default=sa.text("FALSE"), nullable=False),
         sa.Column("corpus_fingerprint", sa.Text(), nullable=False),
@@ -68,8 +73,8 @@ def upgrade() -> None:
     # At most one active run, enforced by the database rather than by every writer remembering to
     # clear the previous one.
     op.create_index(
-        "uq_topic_map_runs_active",
-        "topic_map_runs",
+        "uq_need_maps_active",
+        "need_maps",
         ["active"],
         unique=True,
         postgresql_where=sa.text("active"),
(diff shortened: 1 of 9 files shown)


Bo-Onyx:
@greptile

PR #615, opened 2026-08-04
feat(wiki): AI toolbar on every wiki surface, ChatWidget retired

nmgarza5:
## Description

Replaces the floating ChatWidget with an AI toolbar that docks on every wiki surface, carrying three modes: Chat, Watch, and Launch.

**Chat**
- Full side panel with a history menu grouped into "This Page" and "Recent Chats"
- Per-turn source chips, edit cards, and thinking-state search chips
- Model selector ported from onyx's `ModelSelectorContent` (provider groups, search, checked row)
- Composer accepts **multiple** wiki pages as context, so one turn can span several docs

**Backend**
- `POST /chat/messages` takes `context_paths: list[str]` instead of a single `current_path`
- `GET /chat/sessions?path=` returns `touches_path`, derived from each session's persisted tool calls, so the history menu can group by page with **no schema change**

**Removed**: `ChatWidget`, `ChatWidget.module.css`, `ChatHistoryPanel`.

Reviewed by `code-simplifier` + `comment-analyzer` before push. Fixed from that pass: the armed send CTA built its class without a separator so it rendered unstyled; `message_saved` stamped the last *item* rather than the last *assistant*, silently dropping thumbs on turns ending in a tool call; and the per-path 2048 bound lost in the `current_path` → `context_paths` move (count-only) is restored with a regression test.

The deferred sweeps landed here too: prop-interface extraction, `useSWR` calls moved into `lib/` hooks per `AGENTS.md`, comment corrections, and sessionStorage-backed session continuity across navigation. Retries reuse the persisted user row instead of duplicating it, with regression tests.

## How Has This Been Tested?

Backend: 24 chat-session tests plus the full suite green locally. Frontend: tsc and vitest clean. Continuity, retry dedupe, and the model picker verified live in the browser.

All journeys captured from the running dev app: idle dock, armed composer, thinking state with path chips, streaming with stop, a live wiki edit via chat, model selector groups, history menu with truncation tooltip, side panel with This Page grouping.

## Additional Options


nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

nmgarza5:
@greptile

PR #618, opened 2026-08-04
feat(wiki): brand-new wikis seed the root policy defaults

Bo-Onyx:
Makes the intended org defaults — **Auto Organize on, ingestion Auto Update off (opt-in per scope)** — hold on every new customer deployment, not just environments where someone set the root row by hand.

Mechanism: `seed_if_empty`'s did-seed branch (the one that only runs on the boot that seeds a brand-new wiki) also writes the root (`""`) update-policy row. Properties that fall out of using this seam:

- **Fresh installs only.** A boot over pre-existing content (restored volume, upgrade, migration) stamps the seed marker without touching policy — no existing deployment changes behavior on image pull. Covered by a test.
- **A visible, editable row, not a code default.** A customer who disagrees flips it in the normal Updates panel / API like any folder policy, and the choice sticks — seeding never reruns on a marked database.
- Guarded on "no root row exists", so a bootstrap that somehow raced a policy write never overwrites it.

Pairs with #617 (new-template form defaults Auto Organize on). 5/5 seed tests pass, including the new pre-existing-content guard test.

### DIFF
--- a/backend/app/wiki/seed.py
+++ b/backend/app/wiki/seed.py
@@ -126,8 +126,13 @@ def _seed_root_policy_defaults() -> None:
 
     A visible root policy row rather than a code default, so a customer
     changes it like any folder policy and the choice sticks — this runs
-    only on the boot that seeded the wiki, never on a restored or
-    pre-existing install."""
+    only on the boot that seeds the wiki, never on a restored or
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #624, opened 2026-08-06
feat(ingest): derive the need map from clustered needs

Bo-Onyx:
The producer for the schema #612 landed — one module, `app.ingest.need_map`, that takes the corpus's needs and returns the map.

```
load_needs → cluster_needs → name_cluster → record
             deterministic   one LLM call
             no LLM          per cluster
```

`topics.py` is folded in rather than left beside it: it had exactly one consumer, nothing sits between the two halves, and it was misnamed — it produced *clusters*, not topics, which only exist after the step it didn't do. One module per derived artifact, the way `entity_types.py` is organised.

`cluster_needs` finds which needs *might* belong together, by embedding. It cannot say what they are, and it is imprecise in a specific way: it pulls together needs that merely share a subject area. The widest real cluster held **19 needs across 18 pages**, joined only by being *about Agent Wiki* — a subject, not a facet.

One LLM call per cluster imposes both levels at once:

```
partition by SUBJECT  ->  topics    (splitting what the embedding over-merged)
group by FACET        ->  aspects   (the unit of fan-out)
```

Both from one call because they need the same context — you cannot decide which needs share a facet without first deciding which share a *subject*, and re-supplying the cluster to a second call pays for the same tokens twice.

## Half of this is measured; half runs here for the first time

**Partition is measured.** One call over that 19-need cluster returned 11 topics with 19/19 coverage, including exactly the shape the design assumes:

```
Wiki Auto Management
  requirements and rollout   → 1 page
  architecture and design    → 1 page
  implementation status      → 2 pages     ← fan-out the embedding had buried
```

**Aspect grouping has never run.** It is the untested half I have been flagging since #612, and this is the change that runs it. So the first production derivation is the real test — the prompt asks for aspects to survive their pages maintaining a facet differently (one page a dated log, another a checklist), and whether it does that is not yet knowable from here.

## Trust rules

The map is only worth what its worst response is, so every degradation is chosen rather than incidental. Each is pinned by a test:

| when | outcome | why |
|---|---|---|
| junk entry, out-of-range index, missing topic | that **need** is lost, not the cluster | the call is already paid for; structure beats completeness |
| a need claimed by two aspects | first claim wins | writing it twice double-counts the fan-out, the number this layer is justified by |
| needs left unplaced | absent from the map, logged | a guessed placement is a confident wrong row |
| no needs / no embeddings / nothing named | **records nothing** | each would deactivate a good map and read as "the corpus shares nothing" — indistinguishable from a real finding |

That last row is the one I would most want reviewed. `record()` refuses an empty topic list, but the caller has three earlier ways to reach the same place, and only one of them is an error.

## Cost

One call per cluster, 8-way parallel — the same number entity-type derivation uses, and for the same reason it learned it: run sequentially, 147 pages took ~1.5 hours in production and a pod restart killed it before it recorded anything.

On the measured corpus that is **~150 calls for 238 needs**, most of them single-need clusters whose fan-out is 1. Batching those into shared calls is the obvious optimisation and I have deliberately not done it: it changes what the model sees, and I would rather see real output from the simple shape first than tune a step whose aspect half has never run.

## Refactor carried along

`_complete_json` and `_member_indices` move to `app.ingest.json_completion`. Copying them would have duplicated the truncation handling whose absence silently dropped the eight largest pages of a production wiki — reported as "unparseable JSON", naming neither the page nor the reason. `entity_types` delegates; its 52 tests pass unchanged.

One test change was required and is worth noting: the stub patched `entity_types.client`, which after the move would have stubbed nothing and passed against the real client. It now patches where the call is made.

`StoredNeeds` also gains `content_sha256`, so a derivation can fingerprint its input without re-reading every page.

## Not here: state

Aspect state — the current value a fact gets diffed against — is absent. It changes with every relevant document while the map is a snapshot, and keying it needs an aspect identity that survives re-derivation. Neither exists yet.

Worth flagging the ordering consequence, because it is a real cost: in the eval pipeline the *same* call that groups needs into aspects also reconciles their `current_content` into one merged status. Adding state later means either re-running this call or restructuring it. I judged that acceptable because the aspect half is unproven — but if it works first time, that re-run is the price of having split them.

## Tests

35 in `test_need_map.py` (the clustering half's 16 carried over unchanged), including two end-to-end through `record()` and back out via `aspects_for_page()` — the join no unit test can make, where a field renamed on either side passes everything else and fails here.

Full suite: **2521 passed, 2 skipped**. `ruff` and basedpyright clean.

Task registration is mutation-checked: removing the `run_worker` line fails `test_task_registration`, the guard that caught entity-type derivation being enqueued into a void.

## To run it

`derive_need_map` on the offline queue — deliberately unscheduled, like entity-type derivation, because re-deriving renames topics and ids are stable only within a map.


Bo-Onyx:
Both findings were valid and are fixed in `e9d1a0c`.

**P1 — fractional indices (`json_completion.py`)**

Confirmed and worse than "malformed input dropped": `int(1.5)` is `1`, a perfectly valid position — so the model naming a nonexistent index would silently attach a **different** need to the aspect. Nothing logs, the response still parses, and the map carries a link no model ever asked for.

Dropped rather than truncated. `1.0` is still accepted, since JSON has one number type and that's how an integer arrives — so the rule is "not an integer *value*", not "not an `int`".

Worth noting this behaviour was inherited from `entity_types` when the helper was extracted, so the fix lands there too.

**P2 — fingerprint from a different snapshot (`need_map.py`)**

Also confirmed. `run_derivation` read `page_needs` twice — once to cluster, once to fingerprint — so an extraction landing between them recorded a fingerprint describing a corpus the map wasn't derived from. `load_needs` now takes the rows and the derivation reads once.

The consequence is narrow today only because nothing reads `corpus_fingerprint` yet (noted as a known gap in #612). Storing a wrong one is still worse than storing a right one, since it makes the staleness answer wrong in whichever direction the write fell.

**Tests** — the review was right that neither case was covered. Added, and both mutation-checked:

- restoring truncation → `test_a_fractional_index_is_dropped_not_truncated` fails
- restoring the second read → `test_the_fingerprint_describes_what_was_actually_clustered` fails on the read count

Plus `test_a_whole_float_is_kept` and `test_a_boolean_does_not_claim_the_first_need`, since `True` is an `int` in Python and would otherwise resolve to index 1.

**Also taken:** `log.exception` rather than `log.warning(..., exc_info=True)` in the two `except` blocks, per CLAUDE.md.

Full suite: 2525 passed, 2 skipped. `ruff` and basedpyright clean.

Bo-Onyx:
@greptile

PR #628, opened 2026-08-10
feat(editor): synthetic animated caret (blink + glide + affinity)

raunakab:
## Summary

The native text caret can't be animated — the platform exposes only `caret-color`. So this hides it (on fine pointers) and draws a synthetic replacement (`extensions/caret.ts`) that can **fade-blink** and **glide** on movement.

- **Persistent overlay**, not a `Decoration.widget` (a widget's DOM rebuilds on every reposition, resetting the transition). Lives inside the `.editor-prose` scroller in scroll-origin coordinates, so it scrolls natively and the glide fires only on real caret moves, never on scroll.
- **Blink** is an opacity keyframe (a fade, not hard on/off); solid while moving, resumes blinking when idle.
- **Glide** is a `transform` transition — every move animates, tunable via a single `GLIDE_MS` constant.
- **Affinity bit**: a keydown listener reconstructs the caret affinity the DOM API hides (backward keys → upstream, forward keys / typed char → downstream) and feeds `coordsAtPos(head, side)`, so an ambiguous boundary (soft-wrap / mark edge) paints on the side the user meant. Clicks win via the click x.

Local-only — peers still render via `yCursorPlugin`.

### Edge handling
- Native caret restored during **IME composition** (candidate windows anchor to it).
- **Disabled on touch/coarse pointers** — selection handles, the loupe, and long-press depend on the native caret.
- Renders only for a **collapsed `TextSelection`** (NodeSelection / GapCursor fall through).
- Padding-box border offset subtracted; `prefers-reduced-motion` no-ops the glide.

### Known limitations (documented in code)
- **RTL/bidi affinity** is not handled — the direction→side mapping inverts inside RTL runs, and the browser's embedding-level info isn't exposed to us.
- Positions against a custom node view that can't be measured **hide** the caret rather than misplace it (graceful; deeper `domAtPos` + `Range` fallback deferred).

Sets up a later PR: driving the caret's color from mark context (the "inside a link" signal) hangs off the same `update()`.

## Test plan

- Fine-pointer device: native caret hidden, synthetic caret shows at a collapsed cursor when focused.
- Blink fades; solid while typing/arrowing, resumes when idle.
- Every move glides — typing, arrows, up/down, and large jumps.
- Arrow to a soft-wrap boundary lands the caret on the intended side.
- Range selection → no caret; blur → no caret; scroll → caret stays glued.
- Touch device / `prefers-reduced-motion`: native caret / no glide.
- Peers still render their carets.
- Tune `GLIDE_MS` (`caret.ts`) to change glide speed.

### DIFF
--- a/frontend/src/app/css/editor.css
+++ b/frontend/src/app/css/editor.css
@@ -461,9 +461,13 @@
   /* Above the peer caret (z-index: 1) and positioned image nodes. */
   z-index: 2;
   will-change: transform, opacity;
-  /* Glide between positions; disabled inline for line changes / big jumps. */
-  transition: transform 90ms ease;
-  animation: editor-caret-blink 1.06s ease-in-out infinite;
+  /* What glides and how it eases; *how long* is set inline per move from
+     GLIDE_MS (caret.ts) — hence no duration here, so an un-set move snaps.
+     `transform` must be named explicitly: the initial `all` would drag the
+     inline height and the blink's opacity into the same transition. */
+  transition-property: transform;
+  transition-timing-function: ease;
+  animation: editor-caret-blink 1s ease-in-out infinite;
 }
 /* Solid while moving/typing; the fade-blink resumes once still (caret.ts). */
 .editor-caret.is-moving {
--- a/frontend/src/app/css/editor.css
+++ b/frontend/src/app/css/editor.css
@@ -17,6 +17,11 @@
      write against. */
   height: 100%;
   overflow-y: auto;
+  /* Containing block for the synthetic caret overlay (.editor-caret), which is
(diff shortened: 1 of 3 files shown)


raunakab:
Closing — not pursuing this for now.

The synthetic caret itself (blink, glide, positioning) works, but the affinity half turned into a dead end. Making a soft wrap two caret stops requires owning text navigation outright, because the browser keeps its own affinity bit that we can't read or write, and any bit of ours that disagrees with it breaks vertical motion and line-edge keys. Details and the reasoning are in the `caret.ts` docstring and the commit history on the branch.

Branch is pushed and left intact if anyone wants to pick it up.

PR #630, opened 2026-08-12
fix(coedit): fold a conflicting out-of-band commit into a clean session

Bo-Onyx:
Agent/MCP writes land in git but never reach the editor when the write is a substantial rewrite of a previously co-edited page. The editor renders from the session's stored Yjs snapshot; a large out-of-band rewrite makes the live-rebase fold conflict (by design — the codec-normalized reconstruction differs bytewise from the git base, so a big rewrite overlaps the phantom delta), and the conflict fallback hands off to the checkpoint engine — which no-oped. The page then serves stale content indefinitely. Small edits fold silently, which is why this went unnoticed.

Three defects, three fixes:

1. **`checkpoint_session`'s clean gate** returned early whenever `ydoc_seq == ydoc_checkpointed_seq`, so a session with no local edits — exactly the state a conflicting fold hands off — never reached the merge. The gate now skips only when git agrees: clean *and* `base_sha` is still the page's HEAD.
2. **The no-commit tail** assumed the doc already matched HEAD and advanced the watermark with the stale body — which would have stamped the session synced while editors kept old content. The tail now folds HEAD's content into the doc (same splice/reseed/broadcast path as the diverged-commit case) and advances `base_sha`/snapshot to HEAD.
3. **`advance_checkpoint`'s regression guard** (`ydoc_checkpointed_seq < seq`) silently discarded the equal-seq write a clean fold produces — new snapshot, same watermark. Relaxed to `<=`; a stale checkpoint still can't overwrite a more-advanced one.

Plus a heal path: the periodic scan now also enqueues **clean-but-diverged** sessions (one page-scoped git lookup per active session), so a fold handoff lost to a crash/deploy — including any session already stranded in production today — recovers within a scan interval instead of never.

Tests: end-to-end repro (conflicting rewrite on a clean session, engine + task paths, pinned to a deterministically-conflicting fixture: an all-`1.` ordered list the serializer renumbers), scan-heal case, and the full backend suite (2533 passed).

### DIFF
--- a/backend/app/tasks/coedit_checkpoint.py
+++ b/backend/app/tasks/coedit_checkpoint.py
@@ -113,7 +113,12 @@ def scan_coedit_checkpoints() -> None:
     for sess in coedit.list_active_sessions():
         if sess.ydoc_seq != sess.ydoc_checkpointed_seq:
             continue  # dirty — the seq-based scan owns it
-        if sess.base_sha != wiki_git.head_sha_for_path(sess.path):
+        # None means no commit touches the path (or the git call itself
+        # failed — head_sha_for_path runs check=False): nothing to fold,
+        # and the engine's own gate would no-op, so enqueueing would just
+        # re-run every scan pass forever.
+        head = wiki_git.head_sha_for_path(sess.path)
+        if head is not None and sess.base_sha != head:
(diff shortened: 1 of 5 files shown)


Bo-Onyx:
@greptile

PR #631, opened 2026-08-12
fix(coedit): out-of-band commits fold into the page's document row

Bo-Onyx:
Closes the stale-mirror transplant gap: a page's `wiki_documents` row — the lineage every open transplants from — previously advanced **only on co-edit checkpoints**. A stretch of git-only writes (agent edits, ingestion) left the row holding old content, so the next open transplanted an old revision into the fresh session and relied on the open-time drift fold, which can conflict; the reconciliation debt then rode the checkpoint path. This is the mechanism behind both document-corruption incidents (the older page that came back as 17 copies of a stale revision, and this week's duplication storm during a live multi-editor session — variant copies of blocks unioned in by every checkpoint merge, invisible to the identical-copy and restatement tripwires).

The fix keeps the row honest at the source: `on_wiki_commit` now routes a commit to the open session's fold when one exists (unchanged), **else to a new `rebase_document_row`** that folds the commit into the document row. With no session open the row's own body is the merge base, so the fold is a pure splice onto the existing lineage — no three-way, no conflict case, no lineage reset (reseeding would orphan any disconnected client's retained document, the exact #575-class hazard). Guards:

- active session appears before the task runs → SKIP (the session path owns the page);
- no safe block pairing for the splice → SKIP, open-time fold handles it as before;
- `advance_offline` is a compare-and-swap on `base_sha`, so a checkpoint or move re-mirror that lands concurrently wins and the offline fold yields.

Also corrects the module/model docstrings that still described the table as dual-write/"nothing reads this yet" — opens attach from it, and that stale text actively misdirected this investigation.

Tests: row fold on a no-session commit (content + base + decoded snapshot asserted), active-session skip, trigger routing end-to-end, CAS yield. Full backend suite: 2537 passed.

### DIFF
--- a/backend/app/api/coedit.py
+++ b/backend/app/api/coedit.py
@@ -60,7 +60,7 @@ from app.models.coedit import (
 )
 from app.tasks.coedit_checkpoint import checkpoint_coedit_session_task
 from app.tasks.coedit_rebase import rebase_coedit_session
-from app.wiki import acl, coedit, coedit_channel, coedit_live, git, wiki_documents
+from app.wiki import acl, coedit, coedit_channel, coedit_live, git
 from app.wiki.markdown_yjs import seed_doc_from_markdown
 
 router = APIRouter()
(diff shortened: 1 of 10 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #633, opened 2026-08-17
fix(editor): headings render cramped — fixed base line-height inherited into 2em text

Bo-Onyx:
## Problem

Reported by Duo (via Slack): heading lines in the live editor render crushed together — a wrapped heading overlaps itself.

`.editor-prose` sets `line-height: 1.5rem` — a fixed 24px, not a ratio. Every heading inherits that fixed length, so an h1 at 2em (32px) gets a 24px line box: an effective 0.75 line-height. Dates to the original Tiptap editor styling (#557).

## Fix

- Base line-height is now unitless `1.5`, so descendants inherit a ratio. Body text is unchanged (1rem × 1.5 = the same 24px).
- Headings get an explicit `line-height: 1.2`, the usual display-type treatment — h1 goes from a 24px line box to 38.4px.

Verified on the local stack: computed styles before h1 32px/24px → after 32px/38.4px; p and li unchanged at 16px/24px.

### DIFF
--- a/frontend/src/app/css/editor.css
+++ b/frontend/src/app/css/editor.css
@@ -28,9 +28,7 @@
   color: var(--text-04);
   font-family: var(--font-hanken-grotesk), ui-sans-serif;
   font-size: 1rem;
-  /* Unitless, so larger text inherits a ratio, not a fixed length — a fixed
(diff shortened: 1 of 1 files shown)


Bo-Onyx:
@greptile

PR #634, opened 2026-08-17
fix(coedit): family block ids — never emit duplicate _blockId values to clients

Bo-Onyx:
## Problem — the compounding-duplication storm

Git forensics on the latest duplication incident (a page inflating ~35KB → 573KB in an hour, +1 copy of the same blocks per checkpoint, seconds apart, while the editor rendered clean) pinned the generator to a client/server invariant conflict:

1. A block move put an image paragraph directly under a bullet whose line ends in a markdown hard-break (trailing spaces), no blank line between. On reparse that image line is a lazy continuation of the list — `top_level_block_ranges` folds both doc children into **one** block range.
2. `restamp_block_ids` deliberately assigned both children the **same** positional id (the shared-id design), and `checkpoint_body`'s emit-once guard handled that server-side.
3. But the frontend's `UniqueBlockIdentity` extension enforces the opposite invariant: it **nulls** the `_blockId` of any top-level node duplicating another's. Once the shared id reached a connected client (divergence broadcast / reconnect), the client stripped it.
4. Every subsequent checkpoint then saw an id-less child → "brand-new block" → serialized it **in addition to** the verbatim base-body range that already contains its text. One extra copy committed per checkpoint, compounding for as long as the session stayed dirty — invisible in the editor, unbounded in git.

Reproduced exactly in a unit test (1 → 2 → 3 copies per checkpoint pre-fix).

## Fix

`restamp_block_ids` now stamps children that map onto one reparsed range with **family ids** — `b5`, `b5.1`, `b5.2`, … — so every id on the wire is unique and `UniqueBlockIdentity` never fires. All id-keyed lookups in `markdown_splice` resolve through the family base:

- `checkpoint_body`: range lookup, touch-membership, and the untouched emit-once guard all key on `_family_base` (touching any member re-serializes the whole family, same semantics as before).
- `_splice_table`: row lookups normalize `b5.1:r0` → `b5:r0`, preserving row-level verbatim slicing for family tables.
- Legacy snapshots holding literally-shared ids normalize to the same base and keep their previous (correct) behavior.

Family ids can't collide with parse-derived ids (`b<int>` only) or client-minted ids, and no consumer outside this module interprets the id format (audited backend + frontend). Server-only change — no frontend deploy ordering.

## Tests

- New: the storm scenario end-to-end (seed → move image under a hard-break bullet → checkpoint → restamp → type elsewhere → repeated checkpoints stay at exactly one image; asserts restamp emits no duplicated ids for a client to null).
- New: legacy literally-shared ids still dedup without dropping content.
- Updated: adjacent-lists and adjacent-tables sharing tests to the family scheme (same stability assertions).
- 194 tests across the codec/coedit suites pass; verified a live checkpoint on a local stack commits cleanly.

### DIFF
--- a/backend/app/wiki/markdown_splice.py
+++ b/backend/app/wiki/markdown_splice.py
@@ -33,6 +33,8 @@ Two pieces:
 from __future__ import annotations
 
 import difflib
+import re
+from typing import overload
 
 from pycrdt import Doc, Subscription, XmlElement, XmlFragment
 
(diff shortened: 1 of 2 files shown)


Bo-Onyx:
@greptile

PR #635, opened 2026-08-17
feat(editor): selection toolbar — block style, inline marks, link, Add Comment

Bo-Onyx:
## What

Selecting text in the live editor now opens a formatting menu (per the design mock) where the plain "💬 Comment" pill used to be:

- **Block style switcher** — Normal text / Heading 1–3 / Bullet list / Numbered list / To-do list, with the selection's current style shown and checked.
- **Inline marks** — Bold, Italic, Strikethrough (svg letter glyphs; Opal has no typography icons), Code, and Link (inline URL input; removing reuses the same button). All five round-trip through the markdown codec.
- **Add Comment** — the existing behavior, moved into the menu.

Scope per discussion: Copy Link / Edit with AI / Watch Changes / Launch Agent from the mock are deliberately left out for now.

Read-only surfaces (history versions, no write permission) keep the plain Comment pill — formatting commands need a writable session.

## Underline is disabled (including Cmd+U)

Markdown has no underline, and the codec confirms it: an underline mark serializes to plain text, so Cmd+U rendered formatting that silently vanished on the next checkpoint. StarterKit's underline extension is now off entirely until the codec grows a representation — no button, no shortcut, no silent loss.

## How

- `CoeditorHandle` gains `formatState` / `toggleMark` / `setBlockStyle` / `setLink`, keeping Tiptap types inside `lib/editor` — the host consumes a typed snapshot and re-reads it after each action (actions change marks without moving the selection, so the selection callback alone would leave button states stale).
- `SelectionToolbar` is a pure component; `FileView` owns positioning (same anchor the pill used) and the comment flow.
Screenshot 2026-08-17 at 4 00 41 PM

Verified on the local stack: block detection (selection in a bullet list shows "Bullet list"), bold on/off round-trips through a checkpoint with clean markdown, link input, Cmd+U no-ops, comment flow unchanged.

### DIFF
--- a/frontend/src/components/wiki/SelectionToolbar.tsx
+++ b/frontend/src/components/wiki/SelectionToolbar.tsx
@@ -135,7 +135,7 @@ export function SelectionToolbar({
   };
 
   return (
-    <div className="flex w-56 flex-col p-1">
+    <div className="flex w-48 flex-col">
       <Popover open={styleOpen} onOpenChange={setStyleOpen}>
         <Popover.Trigger asChild>
           <span className="inline-flex w-full">
--- a/frontend/src/components/wiki/SelectionToolbar.tsx
+++ b/frontend/src/components/wiki/SelectionToolbar.tsx
@@ -26,11 +26,17 @@ const BLOCK_LABELS: Record<BlockStyle, string> = {
   h1: "Heading 1",
   h2: "Heading 2",
   h3: "Heading 3",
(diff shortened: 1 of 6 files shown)


Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

Bo-Onyx:
@greptile

PR #636, opened 2026-08-18
feat(editor): Tab indents a plain line into the list above

Bo-Onyx:
## What

Tab on a plain line used to fall through to the browser and throw focus out of the editor (the "can only tab to indent when it's a bullet / numbered list / checkbox" report). Now:

- **A plain line below a list indents into it**: the line becomes an item of the same list kind (bullet / numbered / to-do). Blank-line blocks between are skipped — a blank between bullets still reads as one loose list in markdown. The existing `JoinAdjacentLists` merges the result into the preceding list when directly adjacent, so no shared block ids are ever created. Further Tabs nest through the list's own handling; Shift-Tab lifts as before.
- **Tab anywhere else outside lists/code blocks is swallowed** so focus stays in the editor.
- Inside lists and code blocks, nothing changes — the native handling keeps owning Tab.

This is the only indent markdown can express for a plain line (block nesting outside a list has no markdown form), so it's pure list commands — no codec changes.

## Tests

Vitest through a real headless Tiptap editor with the production list extensions (`MixedTaskList`, `JoinAdjacentLists`, `UniqueBlockIdentity`): direct join, blank-skip join, task-list join, swallow with no list above, and native sink retained inside lists. Also verified live on the local stack: plain line + Tab → joins the list; Tab → nests; Shift-Tab → lifts; committed markdown stays clean.

## Note from live testing (pre-existing, not addressed here)

Keystrokes typed in the first moments after a page loads — before the coedit session finishes its write-ready handshake — are silently dropped (the editor is briefly read-only with no visual indication). Made this PR's verification interesting; filed here for visibility.

### DIFF
--- a/frontend/src/lib/editor/extensions/index.ts
+++ b/frontend/src/lib/editor/extensions/index.ts
@@ -28,6 +28,7 @@ import { CommandMenu } from "@/lib/editor/extensions/commandMenu";
 import { AnchoredHighlights } from "@/lib/editor/extensions/highlights";
 import { imageSupport } from "@/lib/editor/extensions/images";
 import { LinkHover } from "@/lib/editor/extensions/linkHover";
+import { TabIndent } from "@/lib/editor/extensions/tabIndent";
 import { LinkInput } from "@/lib/editor/extensions/linkInput";
 import { presenceExtension } from "@/lib/editor/extensions/presence";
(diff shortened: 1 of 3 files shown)


Bo-Onyx:
@greptile

